PR #76, opened 2025-07-22
Add dotnet-isolated app and move test apps to root

liliankasem:

n/a

### Pull request checklist

### Additional information

- Adding a dni app for testing the worker extension.
- Moving test apps to project root


Directory.Build.props
```diff
@@ -14,6 +14,7 @@
     <EngRoot>$(RepoRoot)eng/</EngRoot>
     <EngBuildRoot>$(EngRoot)build/</EngBuildRoot>
     <EngResourceRoot>$(EngRoot)res/</EngResourceRoot>
+    <SamplesRoot>$(RepoRoot)samples/</SamplesRoot>
```
liliankasem, line 17: @fabiocav we talked about moving the test apps to root last week - let me know if we still want to do that

fabiocav: I'm OK with us doing that, but if we're moving them to `samples`, we should name them accordingly (i.e., no longer "test" apps).

The initial app wasn't initially build as a sample, but as a validation app (doesn't necesarily show different options and patterns). Would make sense to have samples here, but we need to decide if we want to maintain two very similar samples in different repos, as we have a dedicated samples repo with the same scenario.

mattchenderson: Nothing in-proc should be positioned as a sample at all.

liliankasem: It doesn't have to be called samples, the intention is for validation. Did you have a different name you wanted for the folder? I can also leave them in `test/` and just add the isolated app.

liliankasem: Moved back to test/

jviau (approved):

fabiocav (approved):
Minor comments

jviau (approved):

Issue #70, opened 2025-07-02
Optional arguments binding to non-null values.


mattchenderson:
### Version

Unreleased local build of repository (after preview.6)

### Description

When an argument is not required, it may be omitted from the tool call request. This should lead to a null input binding value. However, it seems like strings, for example, will get the empty string.

If you navigate to the argument through the `ToolInvocationContext`, you will successfully get a null value.

I suspect this is a worker-side concern, given that the ToolInvocationContext argument list seems correct.

This feels potentially connected to https://github.com/Azure/azure-functions-dotnet-worker/pull/2155

### Steps to reproduce

1. Create the following function:

  ```csharp
  public class HelloTool(ILogger<HelloTool> logger)
  {
      [Function(nameof(SayHello))]
      public string SayHello(
          [McpToolTrigger(nameof(SayHello), "Responds to the user with a hello message.")] ToolInvocationContext context,
          [McpToolProperty(nameof(name), "string", "The name of the person to greet.")] string? name
      )
      {
          logger.LogInformation("C# MCP tool trigger function processed a request.");
          var entityToGreet = context?.Arguments?.GetValueOrDefault("name") ?? "world";
          return $"Hello, {entityToGreet}! This is an MCP Tool!";
      }
  }
  ```

2. Run this locally, and connect GitHub CoPilot to it.
3. Set a breakpoint on the logger line.
4. Prompt the agent with something like "Say hello using the MCP tool" and allow the tool call to run. Confirm that no parameters are included in the request. This could require more specific prompting, depending on how the agent interprets the default.
5. The breakpoint should be hit. Observe that the value of `name` is the empty string.
6. Continue debugging.
7. The agent will get the response with "Hello, world!" included, which is derived from `context?.Arguments?.GetValueOrDefault("name")` returning `null`.

mattchenderson:
From poking around a bit in a debugging session, it looks like the trick is that the Source is just the empty string itself. This allows [Microsoft.Azure.Functions.Worker.Converters.TypeConverter](https://github.com/Azure/azure-functions-dotnet-worker/blob/ba0f4dc6e0318af9a493198179dc46a29afb7e57/src/DotNetWorker.Core/Converters/TypeConverter.cs) to engage and provide a result. It therefore sets the result value to the source itself, so the empty string.

Issue #59, opened 2025-06-12
Worker serialization error for Boolean tool properties


mattchenderson:
### Version

1.0.0-preview.6

### Description

If a Boolean type is used for a tool property, there will be a runtime error when the tool is called. Here's the example from the original report at https://github.com/Azure-Samples/remote-mcp-functions-dotnet/issues/21 by @MartinM85:

> Error converting 4 input parameters for Function 'GeneratePassword': Cannot convert input parameter 'requireLowercase' to type 'System.Boolean' from type 'System.String'. Error:System.Text.Json.JsonException: 'T' is an invalid start of a value. Path: $ | LineNumber: 0 | BytePositionInLine: 0.
---> System.Text.Json.JsonReaderException: 'T' is an invalid start of a value. LineNumber: 0 | BytePositionInLine: 0.
at System.Text.Json.ThrowHelper.ThrowJsonReaderException(Utf8JsonReader& json, ExceptionResource resource, Byte nextByte, ReadOnlySpan1 bytes) at System.Text.Json.Utf8JsonReader.ConsumeValue(Byte marker) at System.Text.Json.Utf8JsonReader.ReadFirstToken(Byte first) at System.Text.Json.Utf8JsonReader.ReadSingleSegment() at System.Text.Json.Utf8JsonReader.Read() at System.Text.Json.Serialization.JsonConverter1.ReadCore(Utf8JsonReader& reader, T& value, JsonSerializerOptions options, ReadStack& state)
--- End of inner exception stack trace ---
at System.Text.Json.ThrowHelper.ReThrowWithPath(ReadStack& state, JsonReaderException ex)
at System.Text.Json.Serialization.JsonConverter1.ReadCore(Utf8JsonReader& reader, T& value, JsonSerializerOptions options, ReadStack& state) at System.Text.Json.Serialization.Metadata.JsonTypeInfo1.ContinueDeserialize(ReadBufferState& bufferState, JsonReaderState& jsonReaderState, ReadStack& readStack, T& value)
at System.Text.Json.Serialization.Metadata.JsonTypeInfo1.DeserializeAsync(Stream utf8Json, CancellationToken cancellationToken) at System.Text.Json.Serialization.Metadata.JsonTypeInfo1.DeserializeAsObjectAsync(Stream utf8Json, CancellationToken cancellationToken)
[...]

PR #83, opened 2025-07-26
Add ToolInvocationContext to FunctionContext

liliankasem:

resolves #issue_for_this_pr

### Pull request checklist

### Additional information

Additional PR information

### DIFF
--- a/src/Microsoft.Azure.Functions.Worker.Extensions.Mcp/Constants.cs
+++ b/src/Microsoft.Azure.Functions.Worker.Extensions.Mcp/Constants.cs
@@ -1,0 +1,9 @@
+// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License.
+
+namespace Microsoft.Azure.Functions.Worker.Extensions.Mcp;
+
+internal static class Constants
+{
+    public const string ToolInvocationContextKey = "ToolInvocationContext";
+}
--- a/src/Microsoft.Azure.Functions.Worker.Extensions.Mcp/Converters/ToolInvocationContextConverter.cs
+++ b/src/Microsoft.Azure.Functions.Worker.Extensions.Mcp/Converters/ToolInvocationContextConverter.cs
@@ -1,0 +1,29 @@
+// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the MIT License. See License.txt in the project root for license information.
+
+using Microsoft.Azure.Functions.Worker.Converters;
+
+namespace Microsoft.Azure.Functions.Worker.Extensions.Mcp.Converters;
+
+internal class ToolInvocationContextConverter : IInputConverter
+{
+    public ValueTask<ConversionResult> ConvertAsync(ConverterContext context)
+    {
+        ArgumentNullException.ThrowIfNull(context);
+
+        object? target = null;
+
+        if (context.TargetType == typeof(ToolInvocationContext)
+            && context.FunctionContext.TryGetToolInvocationContext(out var toolContext))
+        {
+            target = toolContext;
+        }
+
+        if (target is not null)
+        {
+            return new ValueTask<ConversionResult>(ConversionResult.Success(target));
+        }
+
+        return new ValueTask<ConversionResult>(ConversionResult.Unhandled());
+    }
+}
--- a/src/Microsoft.Azure.Functions.Worker.Extensions.Mcp/DependencyInjection/HostBuilderExtensions.cs
+++ b/src/Microsoft.Azure.Functions.Worker.Extensions.Mcp/DependencyInjection/HostBuilderExtensions.cs
@@ -2,8 +2,12 @@
 // Licensed under the MIT License.
 
 using Microsoft.Azure.Functions.Worker.Core.FunctionMetadata;
+using Microsoft.Azure.Functions.Worker.Extensions.Mcp;
 using Microsoft.Azure.Functions.Worker.Extensions.Mcp.Configuration;
+using Microsoft.Azure.Functions.Worker.Extensions.Mcp.Converters;
 using Microsoft.Azure.Functions.Worker.Extensions.Mcp.DependencyInjection;
+using Microsoft.Extensions.DependencyInjection;
+using Microsoft.Extensions.Hosting;
 
 namespace Microsoft.Azure.Functions.Worker.Builder;
 
@@ -17,6 +21,24 @@ public static class McpHostBuilderExtensions
     public static IFunctionsWorkerApplicationBuilder EnableMcpToolMetadata(this IFunctionsWorkerApplicationBuilder builder)
     {
         builder.Services.Decorate<IFunctionMetadataProvider, McpFunctionMetadataProvider>();
+
+        return builder;
+    }
+
+    // TODO: Better naming
+    public static IFunctionsWorkerApplicationBuilder UseMcpTool(this IFunctionsWorkerApplicationBuilder builder)
+    {
+        if (builder is null)
+        {
+            throw new ArgumentNullException(nameof(builder));
+        }
+
+        builder.UseMiddleware<FunctionsMcpContextMiddleware>();
+
+        builder.Services.Configure<WorkerOptions>(static (workerOption) =>
+        {
+            workerOption.InputConverters.RegisterAt<ToolInvocationContextConverter>(0);
+        });
 
         return builder;
     }
--- a/src/Microsoft.Azure.Functions.Worker.Extensions.Mcp/FunctionsMiddleware/FunctionsMcpContextMiddleware.cs
+++ b/src/Microsoft.Azure.Functions.Worker.Extensions.Mcp/FunctionsMiddleware/FunctionsMcpContextMiddleware.cs
@@ -1,0 +1,30 @@
+// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the MIT License. See License.txt in the project root for license information.
+
+using Microsoft.Azure.Functions.Worker.Middleware;
+
+namespace Microsoft.Azure.Functions.Worker.Extensions.Mcp;
+
+internal class FunctionsMcpContextMiddleware : IFunctionsWorkerMiddleware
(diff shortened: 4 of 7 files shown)


src/Microsoft.Azure.Functions.Worker.Extensions.Mcp/FunctionsMiddleware/FunctionsMcpContextMiddleware.cs
```diff
@@ -0,0 +1,30 @@
...
+
+internal class FunctionsMcpContextMiddleware : IFunctionsWorkerMiddleware
+{
+    public FunctionsMcpContextMiddleware() { }
+
+    public async Task Invoke(FunctionContext context, FunctionExecutionDelegate next)
+    {
+        // At the time that this is invoked, nothing has yet hydrated the ToolInvocationContext from the BindingData.
+        // Are we supposed to hydrate or call the converter from here? (I don't think this makes sense)
+        if (context.BindingContext.BindingData.TryGetValue("mcptoolcontext", out var toolContextValue))
```
liliankasem, line 16: @fabiocav I think I'm missing something here. Is the expectation that we already have a `ToolInvocationContext` hydrated at this stage that we can use/add to the `FunctionContext`? What am I missing to connect the dots?

src/Microsoft.Azure.Functions.Worker.Extensions.Mcp/DependencyInjection/HostBuilderExtensions.cs
```diff
@@ -20,4 +24,22 @@ public static IFunctionsWorkerApplicationBuilder EnableMcpToolMetadata(this IFun
 
         return builder;
     }
-}
\ No newline at end of file
+
+    // TODO: Better naming
```
liliankasem, line 28: Would love some feedback/advice on where to best place this, and naming

src/Microsoft.Azure.Functions.Worker.Extensions.Mcp/Converters/ToolInvocationContextConverter.cs
```diff
@@ -0,0 +1,30 @@
...
+namespace Microsoft.Azure.Functions.Worker.Extensions.Mcp.Converters;
+
+[SupportedTargetType(typeof(ToolInvocationContext))]
+internal class ToolInvocationContextConverter : IInputConverter
+{
+    public ValueTask<ConversionResult> ConvertAsync(ConverterContext context)
+    {
+        ArgumentNullException.ThrowIfNull(context);
+
+        object? target = null;
```
fabiocav, line 15: This shouldn't be required. It's fine to work with the typed variable created in the `TryGetToolInvocationContext` call.

src/Microsoft.Azure.Functions.Worker.Extensions.Mcp/Converters/TargetTypeConversionHelper.cs
```diff
@@ -0,0 +1,52 @@
...
+
+            if (value is IConvertible)
+            {
+                result = Convert.ChangeType(value, underlyingType);
+                return true;
+            }
+
+            var json = JsonSerializer.Serialize(value);
+            result = JsonSerializer.Deserialize(json, underlyingType);
+            return result != null || !underlyingType.IsValueType;
```
liliankasem, line 44:
```suggestion
            return result is not null || !underlyingType.IsValueType;
```

test/TestAppIsolated/Program.cs
```diff
@@ -10,6 +10,6 @@
 //     .AddApplicationInsightsTelemetryWorkerService()
 //     .ConfigureFunctionsApplicationInsights();
 
-builder.EnableMcpToolMetadata();
```
liliankasem, line 13: Removed this as we're registered the decorator in startup, but we can also include this in `ConfigureMcpExtension` instead

src/Microsoft.Azure.Functions.Worker.Extensions.Mcp/DependencyInjection/HostBuilderExtensions.cs
```diff
@@ -20,4 +24,18 @@ public static IFunctionsWorkerApplicationBuilder EnableMcpToolMetadata(this IFun
...
         return builder;
     }
-}
\ No newline at end of file
+
+    public static IFunctionsWorkerApplicationBuilder ConfigureMcpExtension(this IFunctionsWorkerApplicationBuilder builder)
+    {
+        ArgumentNullException.ThrowIfNull(builder);
+
+        builder.UseMiddleware<FunctionsMcpContextMiddleware>();
```
mattchenderson, line 32: Does this need a condition? It looks like this middleware will be run against any invocation. I see we effectively guard against that in `TryGetMcpToolTriggerName`, but I wonder if a `UseWhen` going through the `FunctionDefinition` would be a more efficient circuit break.

src/Microsoft.Azure.Functions.Worker.Extensions.Mcp/McpJsonContext.cs
```diff
@@ -0,0 +1,12 @@
+// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License.
+
+using System.Text.Json.Serialization;
+
+namespace Microsoft.Azure.Functions.Worker.Extensions.Mcp;
+
+[JsonSourceGenerationOptions(WriteIndented = true)]
+[JsonSerializable(typeof(ToolInvocationContext))]
+internal sealed partial class McpJsonContext : JsonSerializerContext
```
mattchenderson, line 10: Where is this used? Is it still intended to be here?

src/Microsoft.Azure.Functions.Worker.Extensions.Mcp/Constants.cs
```diff
@@ -0,0 +1,9 @@
+// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License.
+
+namespace Microsoft.Azure.Functions.Worker.Extensions.Mcp;
+
+internal static class Constants
+{
+    public const string ToolInvocationContextKey = "ToolInvocationContext";
```
mattchenderson, line 8: This is less feedback and more of a question - would something like a `nameof(ToolInvocationContext)` be appropriate for a constant like this? I don't think it actually matters, but curious if something like that would be considered a problem potentially down the line.

mattchenderson (commented):
General question: should this be marked to resolve #70 since #84 did and those changes are also included here?

From testing a build from this branch, I am still seeing the behavior reported in #70, so I don't know that it resolves it yet, but perhaps that's still the intent?

Also am I remembering correctly that this might also be intended to tackle #59?

src/Microsoft.Azure.Functions.Worker.Extensions.Mcp/DependencyInjection/HostBuilderExtensions.cs
```diff
@@ -20,4 +24,18 @@ public static IFunctionsWorkerApplicationBuilder EnableMcpToolMetadata(this IFun
 
         return builder;
     }
-}
\ No newline at end of file
+
+    public static IFunctionsWorkerApplicationBuilder ConfigureMcpExtension(this IFunctionsWorkerApplicationBuilder builder)
```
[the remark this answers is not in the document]
liliankasem: This replaces the `EnableMcpToolMetadata` call that customers need to make today

src/Microsoft.Azure.Functions.Worker.Extensions.Mcp/Converters/JsonConverters/DictionaryStringObjectJsonConverter.cs
```diff
@@ -0,0 +1,54 @@
...
+        JsonValueKind.Null => null,
+        JsonValueKind.Object => ReadObject(element),
+        JsonValueKind.Array => element.EnumerateArray().Select(ReadValue).ToList(),
+        _ => throw new JsonException($"Unsupported JSON token: {element.ValueKind}")
+    };
+
+    public override void Write(Utf8JsonWriter writer, Dictionary<string, object> value, JsonSerializerOptions options)
+    {
+        // Create options without the custom converter to prevent recursion
+        var tempOptions = new JsonSerializerOptions(options);
```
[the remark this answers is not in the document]
liliankasem: Gosh this was many weeks ago so struggling to remember, I think when it was that when this `Write` method just contained `JsonSerializer.Serialize(writer, value, typeof(Dictionary<string, object>)` (without any options), I was getting a stack overflow? Or some other major exception happening.

To fix, I removed the converter so that the custom converter can delegate to the built-in serializer for the actual dictionary serialization, instead of recursively calling itself forever.

src/Microsoft.Azure.Functions.Worker.Extensions.Mcp/DependencyInjection/HostBuilderExtensions.cs
```diff
@@ -20,4 +24,20 @@ public static IFunctionsWorkerApplicationBuilder EnableMcpToolMetadata(this IFun
...
+    {
+        ArgumentNullException.ThrowIfNull(builder);
+
+        builder.EnableMcpToolMetadata();
+
+        builder.UseMiddleware<FunctionsMcpContextMiddleware>();
+
+        builder.Services.Configure<WorkerOptions>(static (workerOption) =>
+        {
+            workerOption.InputConverters.RegisterAt<ToolInvocationContextConverter>(0);
```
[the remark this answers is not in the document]
fabiocav: This is still required in order to support binding to a loose parameter of type `ToolInvocationContext`?

That supports scenarios where the trigger is bound to a POCO and a function argument takes in the context.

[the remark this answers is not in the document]
liliankasem: Ah that's right, I very much forgot aboyt standalone/loose context binding

PR #84, opened 2025-07-28
Implement TypeConverter for worker extension

### LINKED ISSUE #70: Optional arguments binding to non-null values.

liliankasem:

resolves #70

### Pull request checklist

### Additional information

- Implemented a TypeConverter that uses the ToolInvocationContext instead of source
  - The expectation is that there is only argument in the Arguments dictionary for this converter to work
  - I also noticed that we get `JsonElement` types from `ToolInvocationContext` so I implemented a switch to handle those cases too.
- Fixes #70 as we are not longer using the built-in TypeConverter that would return an empty string instead of Unhandled.

Questions

1. Do we want to break this convert our into multiple?
2. Do we need to implement other type converters i.e. DateTime etc.?

### DIFF
--- a/src/Microsoft.Azure.Functions.Worker.Extensions.Mcp/Converters/TypeConverter.cs
+++ b/src/Microsoft.Azure.Functions.Worker.Extensions.Mcp/Converters/TypeConverter.cs
@@ -1,0 +1,70 @@
+// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the MIT License. See License.txt in the project root for license information.
+
+using System.Text.Json;
+using Microsoft.Azure.Functions.Worker.Converters;
+
+namespace Microsoft.Azure.Functions.Worker.Extensions.Mcp.Converters;
+
+internal class TypeConverter : IInputConverter
+{
+    public ValueTask<ConversionResult> ConvertAsync(ConverterContext context)
+    {
+        ArgumentNullException.ThrowIfNull(context);
+
+        // Only expecting a single argument, otherwise use a different converter
+        if (!context.FunctionContext.TryGetToolInvocationContext(out ToolInvocationContext? toolContext)
+            || toolContext.Arguments?.Count != 1)
+        {
+            return new ValueTask<ConversionResult>(ConversionResult.Unhandled());
+        }
+
+        object? argValue = toolContext.Arguments.Values.FirstOrDefault();
+
+        if (argValue is JsonElement jsonElement)
+        {
+            var conversionResult = GetJsonElementConversion(jsonElement, context.TargetType);
+            return new ValueTask<ConversionResult>(conversionResult);
+        }
+
+        if (argValue is not null && context.TargetType.IsAssignableFrom(argValue.GetType()))
+        {
+            return new ValueTask<ConversionResult>(ConversionResult.Success(argValue));
+        }
+
+        return new ValueTask<ConversionResult>(ConversionResult.Unhandled());
+    }
+
+    private ConversionResult GetJsonElementConversion(JsonElement jsonElement, Type targetType) => jsonElement.ValueKind switch
+    {
+        JsonValueKind.String when targetType == typeof(string) =>
+            ConversionResult.Success(jsonElement.GetString()),
+
+        JsonValueKind.Number when targetType == typeof(int) && jsonElement.TryGetInt32(out var i) =>
+            ConversionResult.Success(i),
+
+        JsonValueKind.Number when targetType == typeof(long) && jsonElement.TryGetInt64(out var l) =>
+            ConversionResult.Success(l),
+
+        JsonValueKind.Number when targetType == typeof(float) && jsonElement.TryGetSingle(out var f) =>
+            ConversionResult.Success(f),
+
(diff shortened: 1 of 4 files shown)


test/TestAppIsolated/ToolFunctions.cs
```diff
@@ -15,6 +15,17 @@ public TestFunction(ILogger<TestFunction> logger)
         _logger = logger;
     }
 
+    [Function(nameof(SayHello))]
+    public string SayHello(
+        [McpToolTrigger(nameof(SayHello), "Responds to the user with a hello message.")] ToolInvocationContext context,
+        [McpToolProperty(nameof(name), "string", "The name of the person to greet.")] string? name
```
liliankasem, line 21: Validated that this is null when no arguments are provided

src/Microsoft.Azure.Functions.Worker.Extensions.Mcp/Converters/TypeConverter.cs
```diff
@@ -0,0 +1,70 @@
...
+        ArgumentNullException.ThrowIfNull(context);
+
+        // Only expecting a single argument, otherwise use a different converter
+        if (!context.FunctionContext.TryGetToolInvocationContext(out ToolInvocationContext? toolContext)
+            || toolContext.Arguments?.Count != 1)
+        {
+            return new ValueTask<ConversionResult>(ConversionResult.Unhandled());
+        }
+
+        object? argValue = toolContext.Arguments.Values.FirstOrDefault();
```
mattchenderson, line 22: What does this mean? Why would we limit this to only handling the case of a single argument?

fabiocav: I was about to leave a similar question here. Can you please describe the scenario?

src/Microsoft.Azure.Functions.Worker.Extensions.Mcp/Converters/TypeConverter.cs
```diff
@@ -0,0 +1,70 @@
+// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the MIT License. See License.txt in the project root for license information.
+
+using System.Text.Json;
+using Microsoft.Azure.Functions.Worker.Converters;
+
+namespace Microsoft.Azure.Functions.Worker.Extensions.Mcp.Converters;
+
+internal class TypeConverter : IInputConverter
```
fabiocav, line 9: Added to question from @mattchenderson , but also wanted to leave this one comment here: we should use a more descriptive type name here. `TypeConverter` is a bit generic and we should consider something that more specifically describes this converter performs argument binding/conversion.

src/Microsoft.Azure.Functions.Worker.Extensions.Mcp/McpToolPropertyAttribute.cs
```diff
@@ -3,10 +3,12 @@
 
 using Microsoft.Azure.Functions.Worker.Converters;
 using Microsoft.Azure.Functions.Worker.Extensions.Abstractions;
+using Microsoft.Azure.Functions.Worker.Extensions.Mcp.Converters;
 
 namespace Microsoft.Azure.Functions.Worker.Extensions.Mcp;
 
-[ConverterFallbackBehavior(ConverterFallbackBehavior.Default)]
+[InputConverter(typeof(TypeConverter))]
```
fabiocav, line 10: We also need to make sure this conversion applies to the scenarios currently handled by binding data conversion (e.g., all the properties defined using the bootstrapping model). It's possible that would be a different converter altogether, but the scenario needs to be handled.

PR #105, opened 2025-09-19
Ensuring required properties are validated in MCP extension

### LINKED ISSUE #97: MCP extension does not honor required feilds
### Version

preview.7

### Description

When you mark a tool property as required, the expected behaviour is for the MCP extension to enforce that.

Currently the request still goes through to the worker as a null property.

### Steps to reproduce

1. Run the dni test app `/test/TestAppIsolated`
2. Run mcp inspector & connect to the mcp server
3. Invoke a tool that has a _required_ property, but do not set that property
    - e.g. `SaveSnippet` - provide the content but do not provide a name

**Result:**

The invocation will still go through to the worker/function code with a null value.

Image

**Expected:**

The mcp extension should fail the request early as a required property was not provided.

**Other Info:**

Input schema correctly sets `name` as required:

```json
    {
      "name": "savesnippet",
      "description": "Saves a code snippet into your snippet collection.",
      "inputSchema": {
        "type": "object",
        "properties": {
          "Name": {
            "type": "string",
            "description": "The name of the snippet"
          },
          "Content": {
            "type": "string",
            "description": "The content of the snippet"
          }
        },
        "required": [
          "Name"
        ]
      }
    },
```

Tool call only provides set value:

```json
{
  "method": "tools/call",
  "params": {
    "name": "savesnippet",
    "arguments": {
      "Content": "hello world"
    },
    "_meta": {
      "progressToken": 0
    }
  }
}
```

aishwaryabh:

resolves #97 

### Pull request checklist

### Additional information

This PR adds a check in the MCP inspection to ensure the arguments that are required are actually present

### DIFF
--- a/src/Microsoft.Azure.Functions.Extensions.Mcp/Trigger/McpToolListener.cs
+++ b/src/Microsoft.Azure.Functions.Extensions.Mcp/Trigger/McpToolListener.cs
@@ -3,7 +3,9 @@
 
 using Microsoft.Azure.WebJobs.Host.Executors;
 using Microsoft.Azure.WebJobs.Host.Listeners;
+using ModelContextProtocol;
 using ModelContextProtocol.Protocol;
+using System.Text.Json;
 
 namespace Microsoft.Azure.Functions.Extensions.Mcp;
 
@@ -33,27 +35,24 @@ internal sealed class McpToolListener(ITriggeredFunctionExecutor executor,
 
     public async Task<CallToolResult> RunAsync(CallToolRequestParams callToolRequest, CancellationToken cancellationToken)
     {
-        var execution = new CallToolExecutionContext(callToolRequest);
+        // Validate required properties are present in the incoming request and are not null or empty.
+        ValidateArgumentsHaveRequiredProperties(Properties, callToolRequest);
 
+        var execution = new CallToolExecutionContext(callToolRequest);
         var input = new TriggeredFunctionData
         {
             TriggerValue = execution
         };
-
         var result = await Executor.TryExecuteAsync(input, cancellationToken);
-
         if (!result.Succeeded)
         {
             throw result.Exception;
         }
-
         var toolResult = await execution.ResultTask;
-
         if (toolResult is null)
         {
             return new CallToolResult { Content = [] };
         }
-
         return new CallToolResult
         {
             Content =
@@ -64,7 +63,45 @@ internal sealed class McpToolListener(ITriggeredFunctionExecutor executor,
                 }
             ]
         };
+    }
 
+    internal static void ValidateArgumentsHaveRequiredProperties(ICollection<IMcpToolProperty> properties, CallToolRequestParams callToolRequest)
+    {
+        if (properties is not null && properties.Any(p => p.Required))
+        {
+            var missing = new List<string>();
+            foreach (var prop in properties.Where(p => p.Required))
+            {
+                if (callToolRequest.Arguments == null
+                    || !callToolRequest.Arguments.TryGetValue(prop.PropertyName, out JsonElement value)
+                    || IsValueNullMissingOrEmpty(value))
+                {
+                    missing.Add(prop.PropertyName);
+                    continue;
+                }
+            }
+            if (missing.Count > 0)
+            {
+                // Fail early with an MCP InvalidParams error so the client sees a validation error instead of
+                // the invocation proceeding to the worker with null values.
+                throw new McpException($"Missing required tool properties: {string.Join(", ", missing)}", McpErrorCode.InvalidParams);
+            }
+        }
+    }
 
+    private static bool IsValueNullMissingOrEmpty(JsonElement value)
+    {
+        return value.ValueKind switch
+        {
+            JsonValueKind.Null => true,
+            JsonValueKind.Undefined => true,
+            JsonValueKind.String => string.IsNullOrWhiteSpace(value.GetString()),
+            JsonValueKind.Array => value.GetArrayLength() == 0,
+            JsonValueKind.Object => value.EnumerateObject().Any() == false,
+            JsonValueKind.Number => false, // Numbers can be 0, which is valid
+            JsonValueKind.True => false,   // Boolean true is valid
+            JsonValueKind.False => false,  // Boolean false is valid
+            _ => true
+        };
     }
 }
--- a/test/Extensions.Mcp.Tests/McpToolListenerTests.cs
+++ b/test/Extensions.Mcp.Tests/McpToolListenerTests.cs
@@ -1,0 +1,184 @@
+// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License.
+
+using Microsoft.Azure.WebJobs.Host.Executors;
+using ModelContextProtocol;
+using ModelContextProtocol.Protocol;
+using Moq;
+using System.Text.Json;
+using Xunit;
+
+namespace Microsoft.Azure.Functions.Extensions.Mcp.Tests;
+public class McpToolListenerTests
+{
+    private static IMcpToolProperty CreateProperty(string name, bool required)
+    {
+        var mock = new Mock<IMcpToolProperty>();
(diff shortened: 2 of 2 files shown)


src/Microsoft.Azure.Functions.Extensions.Mcp/Trigger/McpToolListener.cs
```diff
@@ -64,7 +63,45 @@ public async Task<CallToolResult> RunAsync(CallToolRequestParams callToolRequest
...
+                // Fail early with an MCP InvalidParams error so the client sees a validation error instead of
+                // the invocation proceeding to the worker with null values.
+                throw new McpException($"Missing required tool properties: {string.Join(", ", missing)}", McpErrorCode.InvalidParams);
+            }
+        }
+    }
 
+    private static bool IsValueNullMissingOrEmpty(JsonElement value)
+    {
+        return value.ValueKind switch
```
aishwaryabh, line 94: added these types but let me know if I'm missing something

src/Microsoft.Azure.Functions.Extensions.Mcp/Trigger/McpToolListener.cs
```diff
@@ -33,27 +35,24 @@ public void Cancel() { }
...
-        var execution = new CallToolExecutionContext(callToolRequest);
+        // Validate required properties are present in the incoming request and are not null or empty.
+        ValidateArgumentsHaveRequiredProperties(Properties, callToolRequest);
 
+        var execution = new CallToolExecutionContext(callToolRequest);
         var input = new TriggeredFunctionData
         {
             TriggerValue = execution
         };
-
```
liliankasem, line 42: Can we revert all these new line changes - spaces are nice :)

src/Microsoft.Azure.Functions.Extensions.Mcp/Trigger/McpToolListener.cs
```diff
@@ -64,7 +63,45 @@ public async Task<CallToolResult> RunAsync(CallToolRequestParams callToolRequest
...
+        {
+            var missing = new List<string>();
+            foreach (var prop in properties.Where(p => p.Required))
+            {
+                if (callToolRequest.Arguments == null
+                    || !callToolRequest.Arguments.TryGetValue(prop.PropertyName, out JsonElement value)
+                    || IsValueNullMissingOrEmpty(value))
+                {
+                    missing.Add(prop.PropertyName);
+                    continue;
```
liliankasem, line 80: Shouldn't need to continue here

src/Microsoft.Azure.Functions.Extensions.Mcp/Trigger/McpToolListener.cs
```diff
@@ -64,7 +63,45 @@ public async Task<CallToolResult> RunAsync(CallToolRequestParams callToolRequest
...
+        {
+            JsonValueKind.Null => true,
+            JsonValueKind.Undefined => true,
+            JsonValueKind.String => string.IsNullOrWhiteSpace(value.GetString()),
+            JsonValueKind.Array => value.GetArrayLength() == 0,
+            JsonValueKind.Object => value.EnumerateObject().Any() == false,
+            JsonValueKind.Number => false, // Numbers can be 0, which is valid
+            JsonValueKind.True => false,   // Boolean true is valid
+            JsonValueKind.False => false,  // Boolean false is valid
+            _ => true
```
liliankasem, line 104: Should we add `JsonValueKind.None => true, ` ?

src/Microsoft.Azure.Functions.Extensions.Mcp/Trigger/McpToolListener.cs
```diff
@@ -64,7 +63,45 @@ public async Task<CallToolResult> RunAsync(CallToolRequestParams callToolRequest
...
 
+    private static bool IsValueNullMissingOrEmpty(JsonElement value)
+    {
+        return value.ValueKind switch
+        {
+            JsonValueKind.Null => true,
+            JsonValueKind.Undefined => true,
+            JsonValueKind.String => string.IsNullOrWhiteSpace(value.GetString()),
+            JsonValueKind.Array => value.GetArrayLength() == 0,
+            JsonValueKind.Object => value.EnumerateObject().Any() == false,
```
liliankasem, line 100: I think `!value.EnumerateObject().MoveNext()` is more efficient

Same question as array - does an intentionally empty object count as missing?

src/Microsoft.Azure.Functions.Extensions.Mcp/Trigger/McpToolListener.cs
```diff
@@ -64,7 +63,45 @@ public async Task<CallToolResult> RunAsync(CallToolRequestParams callToolRequest
...
+    }
 
+    private static bool IsValueNullMissingOrEmpty(JsonElement value)
+    {
+        return value.ValueKind switch
+        {
+            JsonValueKind.Null => true,
+            JsonValueKind.Undefined => true,
+            JsonValueKind.String => string.IsNullOrWhiteSpace(value.GetString()),
+            JsonValueKind.Array => value.GetArrayLength() == 0,
```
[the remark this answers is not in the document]
mattchenderson: An empty array is still an array being provided. That is semantically different from the entire property being omitted.

src/Microsoft.Azure.Functions.Extensions.Mcp/Trigger/McpToolListener.cs
aishwaryabh: I don't think there's a JsonValueKind.None value?

src/Microsoft.Azure.Functions.Extensions.Mcp/Trigger/McpToolListener.cs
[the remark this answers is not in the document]
aishwaryabh: Updated so that we only throw an error if JsonKindValue.Kind is Null or Undefined

src/Microsoft.Azure.Functions.Extensions.Mcp/Trigger/McpToolListener.cs
aishwaryabh: Updated so that we only throw an error if JsonKindValue.Kind is Null or Undefined

src/Microsoft.Azure.Functions.Extensions.Mcp/Trigger/McpToolListener.cs
```diff
@@ -64,7 +69,46 @@ public async Task<CallToolResult> RunAsync(CallToolRequestParams callToolRequest
...
+        foreach (var prop in requiredProps)
+        {
+            if (args == null
+                || !args.TryGetValue(prop.PropertyName, out JsonElement value)
+                || IsValueNullOrUndefined(value))
+            {
+                missing.Add(prop.PropertyName);
+            }
+        }
+        if (missing.Count > 0)
```
liliankasem, line 97: nit: new line between branches
```suggestion

        if (missing.Count > 0)
```

src/Microsoft.Azure.Functions.Extensions.Mcp/Trigger/McpToolListener.cs
```diff
@@ -64,7 +69,47 @@ public async Task<CallToolResult> RunAsync(CallToolRequestParams callToolRequest
...
 
+        if (requiredProps is null || requiredProps.Length == 0)
+        {
+            return;
+        }
+
+        var missing = new List<string>();
+        var args = callToolRequest?.Arguments;
+
+        foreach (var prop in requiredProps)
```
liliankasem, line 88: Just spoke to Fabio about this and he had a good point: the shortcircut on lines 80 doesn't help us as the array assignment on line 76 already iterates through the entire properties list to get check of `Required`. So with the foreach here we end up iterating twice.

So the suggestion here is to remove lines 76-78; revert back to checking properties null/length on line 80, and update this foreach loop to check for prop.Required and continue if false

PR #107, opened 2025-09-22
Support raw JSON-RPC messages from workers: Extension changes

### LINKED ISSUE #26: Support raw messages from workers
The extension should support raw JSON-RPC messages from the workers. Support for raw content (what is in place today) should still be supported.

In order to pass a raw JSON-RPC message to the extension, worker implementations must return a payload the complies with the following JSON format:

``` json
{
  "isFunctionsMcpResult": true,
  "content": <JSON-RPC message payload>
}
```

aishwaryabh:

resolves #26

### Pull request checklist

### Additional information

Adding support to the extension to support the following format:
```
{
  "isFunctionsMcpResult": true,
  "content": <JSON-RPC message payload>
}
```


src/Microsoft.Azure.Functions.Extensions.Mcp/Http/SseRequestHandler.cs
```diff
@@ -105,7 +105,7 @@ public async Task HandleMessageRequestAsync(HttpContext context, McpOptions mcpO
             return;
         }
 
-        var message = await context.Request.ReadFromJsonAsync<JsonRpcMessage>(McpJsonSerializerOptions.DefaultOptions, context.RequestAborted);
+        var message = await McpHttpUtility.ExtractJsonRpcMessageSseAsync(context.Request, McpJsonSerializerOptions.DefaultOptions, context.RequestAborted);
```
liliankasem, line 108: You could implement this as a `HttpRequest` extension

src/Microsoft.Azure.Functions.Extensions.Mcp/Http/McpHttpUtility.cs
```diff
@@ -39,4 +42,82 @@ internal static void SetSseContext(HttpContext context)
...
+        // wrapped payloads with the shape: { "isFunctionsMcpResult": true, "content": <JSON-RPC message> }
+        // 
+        // When unwrapOnly is false: If the wrapper is present, deserialize the inner "content" as the JsonRpcMessage. 
+        // Otherwise, deserialize the root object directly as a JsonRpcMessage and return it.
+        //
+        // When unwrapOnly is true: If the wrapper is present, replace the request.Body stream with a memory stream 
+        // containing only the inner content. If no wrapper is present, leave the original body intact.
+
+        // If the body is empty, return null.
+        if (request.ContentLength == null || request.ContentLength == 0)
```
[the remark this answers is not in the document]
liliankasem: We can also use CanRead

src/Microsoft.Azure.Functions.Extensions.Mcp/Http/McpHttpUtility.cs
```diff
@@ -39,4 +42,82 @@ internal static void SetSseContext(HttpContext context)
...
+        {
+            return null;
+        }
+
+        // Read the request body into a JsonDocument for inspection.
+        request.EnableBuffering();
```
liliankasem, line 64: Is this the first point of entry, could this have been set before we got to this method (middleware perhaps?) I don't think there is an issue setting this twice, but I wonder where is the right place

liliankasem, line 64: Could another component have read the body before we get to this part of the code? If so, do we need to reset the position?

src/Microsoft.Azure.Functions.Extensions.Mcp/Http/McpHttpUtility.cs
```diff
@@ -39,4 +42,82 @@ internal static void SetSseContext(HttpContext context)
...
+        try
+        {
+            using var doc = await JsonDocument.ParseAsync(request.Body, cancellationToken: cancellationToken);
+            var root = doc.RootElement;
+
+            JsonElement messageElement = root;
+            bool isWrapped = false;
+
+            if (root.ValueKind == JsonValueKind.Object &&
```
liliankasem, line 73:
```csharp
bool isWrapped =
        root.ValueKind == JsonValueKind.Object &&
        root.TryGetProperty("isFunctionsMcpResult", out var marker) &&
        marker.ValueKind == JsonValueKind.True &&
        root.TryGetProperty("content", out var messageElement);
 ```
 
 TryGetProperty is already setting a conent var  so we don't really need to assign it again

File: src/Microsoft.Azure.Functions.Extensions.Mcp/Http/McpHttpUtility.cs [hunk 5]
```diff
@@ -39,4 +42,82 @@ internal static void SetSseContext(HttpContext context)
...
+                {
+                    var inner = messageElement.GetRawText();
+                    var bytes = System.Text.Encoding.UTF8.GetBytes(inner);
+                    request.Body = new MemoryStream(bytes);
+                    request.ContentLength = bytes.Length;
+                }
+                else
+                {
+                    // Reset position so downstream readers can consume the original body.
+                    request.Body.Seek(0, SeekOrigin.Begin);
```
liliankasem, line 94: For extra safety, maybe check CanSeek before doing this

src/Microsoft.Azure.Functions.Extensions.Mcp/Http/McpHttpUtility.cs
```diff
@@ -39,4 +42,82 @@ internal static void SetSseContext(HttpContext context)
...
+                }
+                return null;
+            }
+            else
+            {
+                var raw = messageElement.GetRawText();
+                return JsonSerializer.Deserialize<JsonRpcMessage>(raw, options);
```
liliankasem, line 101: What are we doing on json exceptions? And do we know that messageElement is json?

src/Microsoft.Azure.Functions.Extensions.Mcp/Http/McpHttpUtility.cs
liliankasem: It might be safer to always reset

src/Microsoft.Azure.Functions.Extensions.Mcp/Http/McpHttpUtility.cs
liliankasem: Not sure that `root.TryGetProperty("content", out var messageElement);` needs to be connected to the result of `isFunctionsMcpResult`?

PR #109, opened 2025-09-27
Flowing transport and client context to tool invocation.

### LINKED ISSUE #20: Context enrichment to expose client and transport details
Currently, we don't expose information about the client, session, or transport. In some cases, users want to process information from the transport, such as request headers.

fabiocav:
Flowing transport and client metadata to tool invocation.

Improvements to transport context handling, refactoring tool invocation context usage, and updating interfaces for consistency. 

resolves #20 

### Pull request checklist


src/Microsoft.Azure.Functions.Worker.Extensions.Mcp/Abstractions/ToolInvocationContext.cs
liliankasem, line 12: Does ToolInvocationContext need to be defined in the core MCP ext AND the worker? Can't we use the one from the core ext? I don't think we're doing anything different?

fabiocav: I was anticipating the question :)

I'm intentionally keeping those separate for now. What we now have from the extension is simply a DTO and they may evolve separately. I should move forward and also make that internal, but some of the changes we've been discussing will likely impact the shape of that type here.

fabiocav: Made some updates to the type (including making it internal)

PR #130, opened 2025-10-22
WebJobs Extension: Implement support for multiple MCP tool result types

### LINKED ISSUE #113: WebJobx Extension: Handle other tool content types + raw mcp json
Update the WebJobs MCP extension to support raw content + other content types beyond just Text

liliankasem:

resolves #113

### Pull request checklist

### Additional information

Update the webjobs extension to support other tool result types beyond just text content blocks.

### DIFF
--- a/test/Extensions.Mcp.Tests/ToolReturnValueBinderTests.cs
+++ b/test/Extensions.Mcp.Tests/ToolReturnValueBinderTests.cs
@@ -118,8 +118,13 @@ public class ToolReturnValueBinderTests
         var context = CallToolExecutionContextHelper.CreateExecutionContext();
         var binder = new ToolReturnValueBinder(context);
 
-        // Bug in MCP SDK: Size must be set to deserialize correctly
-        var innerBlock = new ResourceLinkBlock { Uri = "https://example.com", Name = "Example", Size = 1234};
+        var innerBlock = new ResourceLinkBlock
+        {
+            Uri = "https://example.com",
+            Name = "Example",
+            Size = 1234
+        };
+
         var mcpToolResult = new McpToolResult
         {
             Type = "resource_link",
@@ -143,25 +148,12 @@ public class ToolReturnValueBinderTests
         var context = CallToolExecutionContextHelper.CreateExecutionContext();
         var binder = new ToolReturnValueBinder(context);
 
-        var text = new TextContentBlock { Text = "hello" };
-        var image = new ImageContentBlock
-        {
-            Data = Convert.ToBase64String(new byte[] { 1, 2, 3 }),
-            MimeType = "image/png"
-        };
-        var link = new ResourceLinkBlock
-        {
-            Uri = "https://example.com/resource",
-            Name = "Test Resource",
-            Description = "A test resource for validation",
-            MimeType = "text/plain",
-            Size = 1024
-        };
+        string contentBlocksJson = "[{\"type\":\"text\",\"text\":\"hello\"},{\"type\":\"image\",\"data\":\"AQID\",\"mimeType\":\"image/png\"},{\"type\":\"resource_link\",\"uri\":\"https://example.com/resource\",\"name\":\"Test Resource\"}]";
 
         var mcpToolResult = new McpToolResult
         {
             Type = "multi_content_result",
-            Content = JsonSerializer.Serialize(new object[] { text, image, link })
+            Content = contentBlocksJson
         };
         var json = JsonSerializer.Serialize(mcpToolResult);
 
@@ -179,16 +171,15 @@ public class ToolReturnValueBinderTests
 
         var textBlock = (TextContentBlock)result.Content[0];
         Assert.Equal("text", textBlock.Type);
-        Assert.Equal(text.Text, textBlock.Text);
+        Assert.Equal("hello", textBlock.Text);
 
         var imageBlock = (ImageContentBlock)result.Content[1];
-        Assert.Equal(image.MimeType, imageBlock.MimeType);
-        Assert.Equal(image.Data, imageBlock.Data);
+        Assert.Equal("image/png", imageBlock.MimeType);
+        Assert.Equal("AQID", imageBlock.Data);
 
         var linkBlock = (ResourceLinkBlock)result.Content[2];
-        Assert.Equal(link.Uri, linkBlock.Uri);
-        Assert.Equal(link.MimeType, linkBlock.MimeType);
-        Assert.Equal(link.Name, linkBlock.Name);
+        Assert.Equal("https://example.com/resource", linkBlock.Uri);
+        Assert.Equal("Test Resource", linkBlock.Name);
     }
 
     [Fact]


src/Microsoft.Azure.Functions.Extensions.Mcp/Trigger/McpToolListener.cs
liliankasem: I left this in here as the return value binders will return null if value is null. We could alternatively move this logic to the value binders.

src/Microsoft.Azure.Functions.Extensions.Mcp/Trigger/McpToolTriggerRichContentReturnValueBinder.cs
```diff
@@ -0,0 +1,91 @@
...
+            executionContext.SetResult(rawContentResult);
+            return Task.CompletedTask;
+        }
+
+        ContentBlock? content;
+        try
+        {
+            content = mcpToolResult.Type switch
+            {
+                McpContentToolType.Text => JsonSerializer.Deserialize<TextContentBlock>(mcpToolResult.Content),
```
liliankasem: Here we are expecting the incoming content to be in the mcp tool content block schema format

src/Microsoft.Azure.Functions.Extensions.Mcp/Trigger/McpToolTriggerRichContentReturnValueBinder.cs
```diff
@@ -0,0 +1,91 @@
...
+        var mcpToolResult = JsonSerializer.Deserialize<McpToolResult>(jsonString, McpJsonSerializerOptions.DefaultOptions)
+                    ?? throw new InvalidOperationException("The function return value could not be deserialized to a valid McpToolResult.");
+
+        if (string.IsNullOrWhiteSpace(mcpToolResult.Content))
+        {
+            throw new InvalidOperationException("McpToolResult is missing content.");
+        }
+
+        // Do we want to handle raw content type differently?
+        // Or should it just be a TextContentBlock with Text set to the raw content?
```
fabiocav: As discussed offline; let's leave raw out this PR while we iterate on the appropriate behavior for this.

src/Microsoft.Azure.Functions.Extensions.Mcp/Trigger/McpToolTriggerRichContentReturnValueBinder.cs
```diff
@@ -0,0 +1,91 @@
...
+
+        if (value is not string jsonString)
+        {
+            throw new ArgumentException("Expected JSON string.", nameof(value));
+        }
+
+        var mcpToolResult = JsonSerializer.Deserialize<McpToolResult>(jsonString, McpJsonSerializerOptions.DefaultOptions)
+                    ?? throw new InvalidOperationException("The function return value could not be deserialized to a valid McpToolResult.");
+
+        if (string.IsNullOrWhiteSpace(mcpToolResult.Content))
```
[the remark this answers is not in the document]
liliankasem: I was aiming for lines 17-21 to handle empty responses (when the value is empty), but I see what you're saying about potentialy having an empty **content** block still be valid. In this case, the deseralization below should be able to handle that? Though I would except that to throw if content is null

src/Microsoft.Azure.Functions.Extensions.Mcp/Trigger/McpToolTriggerRichContentReturnValueBinder.cs
```diff
@@ -0,0 +1,76 @@
...
+// Licensed under the MIT License. See License.txt in the project root for license information.
+
+using System.Text.Json;
+using Microsoft.Azure.Functions.Extensions.Mcp.Serialization;
+using Microsoft.Azure.WebJobs.Host.Bindings;
+using ModelContextProtocol.Protocol;
+
+namespace Microsoft.Azure.Functions.Extensions.Mcp;
+
+internal class McpToolTriggerRichContentReturnValueBinder(CallToolExecutionContext executionContext) : IValueBinder
```
liliankasem: This file name is so long 😅

fabiocav: Seal this class.

src/Microsoft.Azure.Functions.Extensions.Mcp/Trigger/McpToolTriggerReturnValueBinder.cs
```diff
@@ -0,0 +1,39 @@
+// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the MIT License. See License.txt in the project root for license information.
+
+using Microsoft.Azure.WebJobs.Host.Bindings;
+using ModelContextProtocol.Protocol;
+
+namespace Microsoft.Azure.Functions.Extensions.Mcp;
+
+internal class McpToolTriggerReturnValueBinder(CallToolExecutionContext executionContext) : IValueBinder
```
fabiocav: Seal this class.

src/Microsoft.Azure.Functions.Extensions.Mcp/Trigger/McpToolTriggerRichContentReturnValueBinder.cs
```diff
@@ -0,0 +1,76 @@
...
+
+        var mcpToolResult = JsonSerializer.Deserialize<McpToolResult>(jsonString, McpJsonSerializerOptions.DefaultOptions)
+                    ?? throw new InvalidOperationException("The function return value could not be deserialized to a valid McpToolResult.");
+
+        if (mcpToolResult.Content is null)
+        {
+            throw new ArgumentNullException(mcpToolResult.Content, "The McpToolResult content cannot be null.");
+        }
+
+        ContentBlock? content;
```
fabiocav: Content here should be a collection to enable multiple results (e.g., different content types, batch results, multi-output formats, etc.). That will match the expectation from `CallToolResult` (you won't have to create a single element collection when this is done).

The logic here would need to treat the raw content as a potential collection. You should consider making that explicit in the contract, where the `Type` would indicate multiple content entries (e.g., `multi-content-result`). That means you'd be able to keep tour existing handling for the types you support (`audio`, `image`, etc.), but you'd need to be able to reuse this when handling a collection of content items, which is a good change as it would also simplify/reduce some of the mapping code.

liliankasem: Addressed this feedback, let me know what you think!

src/Microsoft.Azure.Functions.Extensions.Mcp/McpConstants.cs
```diff
@@ -12,4 +12,14 @@ internal sealed class McpConstants
     public const string MessageEndpoint = "message";
     public const string McpSessionIdHeaderName = "Mcp-Session-Id";
     public const string McpTransportName = "functions_mcp:transportname";
+
+    internal class ToolResultContentTypes
```
fabiocav: seal

src/Microsoft.Azure.Functions.Extensions.Mcp/Trigger/ToolReturnValueBinder.cs
```diff
@@ -0,0 +1,166 @@
...
+        {
+            throw new ArgumentException("Expected JSON string.", nameof(value));
+        }
+
+        var result = JsonSerializer.Deserialize<McpToolResult>(jsonString, McpJsonSerializerOptions.DefaultOptions)
+                     ?? throw new InvalidOperationException("The function return value could not be deserialized to a valid McpToolResult.");
+
+        if (result.Content is null)
+        {
+            throw new ArgumentNullException(nameof(result.Content), "The McpToolResult 'Content' cannot be null.");
```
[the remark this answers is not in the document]
liliankasem: Very fair, I originally was checking null or empty and changed to just null after the first iteration of feedback. I guess I worry that null content will break somewhere down the line so will need to figure out if and where to check for that

src/Microsoft.Azure.Functions.Extensions.Mcp/Trigger/McpToolResult.cs
```diff
@@ -0,0 +1,21 @@
+// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the MIT License. See License.txt in the project root for license information.
+
+namespace Microsoft.Azure.Functions.Extensions.Mcp;
+
+/// <summary>
+/// Represents the result returned by an MCP tool execution.
+/// Utilized when rich content type support is enabled.
+/// </summary>
+public class McpToolResult
```
fabiocav: seal this type

src/Microsoft.Azure.Functions.Extensions.Mcp/Trigger/ToolReturnValueBinder.cs
```diff
@@ -0,0 +1,105 @@
...
+
+            foreach (var item in collection)
+            {
+                blocks.Add(item);
+            }
+        }
+        else
+        {
+            // Otherwise, handle single content block
+            var contentBlock = result.Type switch
```
fabiocav: `ContentBlock` provides a JSON converter that will handle conversion to the appropriate types based on the `Type` value.

The following should be sufficient, and will avoid the need to switch on the type or have the `DeserializeContentBlock` method:

``` csharp
var contentBlock = JsonSerializer.Deserialize<ContentBlock>(result.Content!, McpJsonUtilities.DefaultOptions)
    ?? throw new InvalidOperationException("Failed to deserialize content block."
```

src/Microsoft.Azure.Functions.Extensions.Mcp/Trigger/ToolReturnValueBinder.cs
```diff
@@ -0,0 +1,105 @@
...
+
+    private static Collection<ContentBlock> DeserializeToContentBlockCollection(McpToolResult result)
+    {
+        var blocks = new Collection<ContentBlock>();
+
+        // Explicit multi-content contract
+        if (string.Equals(result.Type, McpConstants.ToolResultContentTypes.MultiContentResult, StringComparison.OrdinalIgnoreCase))
+        {
+
+            var collection = JsonSerializer.Deserialize<Collection<ContentBlock>>(result.Content!, McpJsonSerializerOptions.DefaultOptions)
```
fabiocav: Use `McpJsonUtilities.DefaultOptions` here instead.

PR #136, opened 2025-11-01
Support worker generated input schema via `UseWorkerInputSchema` flag

### LINKED ISSUE #135: Extension changes for supporting richer types
- MCP Extension changes with new bool property called UseWorkerInputSchema
- Validation of required properties should be done with input schema if it is present

aishwaryabh:

resolves #135 

### Pull request checklist

### Additional information
This PR covers the MCP Extension changes with new bool property called `UseWorkerInputSchema` and performs the validation of required properties with input schema if it is present

### DIFF
--- a/src/Microsoft.Azure.Functions.Extensions.Mcp/ToolInputSchema/PropertyBasedToolInputSchema.cs
+++ b/src/Microsoft.Azure.Functions.Extensions.Mcp/ToolInputSchema/PropertyBasedToolInputSchema.cs
@@ -11,14 +11,16 @@ namespace Microsoft.Azure.Functions.Extensions.Mcp.Validation;
 internal sealed class PropertyBasedToolInputSchema : ToolInputSchema
 {
     private readonly ICollection<IMcpToolProperty> _properties;
+    private readonly Lazy<JsonElement> _cachedSchemaElement;
 
     /// <summary>
     /// Initializes a new instance of the PropertyBasedToolInputSchema class.
     /// </summary>
     /// <param name="properties">The tool properties to use for validation.</param>
-    public PropertyBasedToolInputSchema(ICollection<IMcpToolProperty> properties)
+    public PropertyBasedToolInputSchema(IEnumerable<IMcpToolProperty> properties)
     {
-        _properties = properties ?? throw new ArgumentNullException(nameof(properties));
+        _properties = properties.ToList() ?? throw new ArgumentNullException(nameof(properties));
+        _cachedSchemaElement = new Lazy<JsonElement>(BuildSchemaElement);
     }
 
     /// <summary>
@@ -27,6 +29,23 @@ internal sealed class PropertyBasedToolInputSchema : ToolInputSchema
     /// </summary>
     /// <returns>A JsonElement representing the input schema.</returns>
     public override JsonElement GetSchemaElement()
+    {
+        return _cachedSchemaElement.Value;
+    }
+
+    /// <summary>
+    /// Gets the list of required property names for validation.
+    /// </summary>
+    /// <returns>A collection of required property names.</returns>
+    protected override IReadOnlyCollection<string> GetRequiredProperties()
+    {
+        return _properties
+            .Where(p => p.IsRequired)
+            .Select(p => p.PropertyName)
+            .ToList();
+    }
+
+    private JsonElement BuildSchemaElement()
     {
         var props = _properties
             .Where(p => p is not null && !string.IsNullOrWhiteSpace(p.PropertyName))
@@ -88,20 +107,9 @@ internal sealed class PropertyBasedToolInputSchema : ToolInputSchema
         }
 
         ms.Position = 0;
+
         using var doc = JsonDocument.Parse(ms);
         return doc.RootElement.Clone();
-    }
-
-    /// <summary>
-    /// Gets the list of required property names for validation.
-    /// </summary>
-    /// <returns>A collection of required property names.</returns>
-    protected override IReadOnlyCollection<string> GetRequiredProperties()
-    {
-        return _properties
-            .Where(p => p.IsRequired)
-            .Select(p => p.PropertyName)
-            .ToList();
     }
 
     private static void WriteTypeAndEnum(Utf8JsonWriter writer, string propertyType, IReadOnlyList<string> enumValues)
--- a/src/Microsoft.Azure.Functions.Extensions.Mcp/ToolInputSchema/JsonSchemaToolInputSchema.cs
+++ b/src/Microsoft.Azure.Functions.Extensions.Mcp/ToolInputSchema/JsonSchemaToolInputSchema.cs
@@ -2,6 +2,7 @@
 // Licensed under the MIT License.
 
 using System.Text.Json;
+using Microsoft.Azure.Functions.Extensions.Mcp;
 
 namespace Microsoft.Azure.Functions.Extensions.Mcp.Validation;
 
@@ -11,6 +12,7 @@ namespace Microsoft.Azure.Functions.Extensions.Mcp.Validation;
 internal sealed class JsonSchemaToolInputSchema : ToolInputSchema, IDisposable
 {
     private readonly JsonDocument _inputSchema;
+    private readonly Lazy<JsonElement> _cachedSchemaElement;
 
     /// <summary>
     /// Initializes a new instance of the JsonSchemaToolInputSchema class.
@@ -19,6 +21,7 @@ internal sealed class JsonSchemaToolInputSchema : ToolInputSchema, IDisposable
     public JsonSchemaToolInputSchema(JsonDocument inputSchema)
     {
         _inputSchema = inputSchema ?? throw new ArgumentNullException(nameof(inputSchema));
+        _cachedSchemaElement = new Lazy<JsonElement>(() => _inputSchema.RootElement);
     }
 
     /// <summary>
@@ -28,7 +31,7 @@ internal sealed class JsonSchemaToolInputSchema : ToolInputSchema, IDisposable
     /// <returns>A JsonElement representing the input schema.</returns>
     public override JsonElement GetSchemaElement()
     {
-        return _inputSchema.RootElement;
+        return _cachedSchemaElement.Value;
     }
 
     /// <summary>


src/Microsoft.Azure.Functions.Extensions.Mcp/Abstractions/IMcpTool.cs
```diff
@@ -14,5 +15,7 @@ internal interface IMcpTool
 
     public ICollection<IMcpToolProperty> Properties { get; set; }
 
+    public McpInputSchema? InputSchema { get; set; }
```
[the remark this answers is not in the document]
liliankasem: But also let's discuss in the design doc. If we have access to the MCP SDK in the worker, could we not just return a Tool? Without having to do any parsing?

src/Microsoft.Azure.Functions.Extensions.Mcp/Abstractions/IMcpTool.cs
```diff
@@ -14,5 +16,8 @@ internal interface IMcpTool
 
     public ICollection<IMcpToolProperty> Properties { get; set; }
 
+    [JsonPropertyName("inputSchema")]
```
aishwaryabh: not sure if we need this but just added it just in case

src/Microsoft.Azure.Functions.Extensions.Mcp/Trigger/McpToolTriggerBinding.cs
```diff
@@ -132,16 +132,22 @@ private Transport GetTransportInformation(CallToolExecutionContext context)
...
-        var toolProperties = GetProperties(_toolAttribute, _triggerParameter);
+        IList<IMcpToolProperty> toolProperties = [];
 
-        var listener = new McpToolListener(context.Executor, context.Descriptor.ShortName, _toolAttribute.ToolName, _toolAttribute.Description, toolProperties);
+        // Generate tool properties only if input schema was not generated in the worker
+        if (!_toolAttribute.UseWorkerInputSchema)
+        {
+            toolProperties = GetProperties(_toolAttribute, _triggerParameter);
+        }
+
```
liliankasem: We should guard against UseWorkerInputSchema being true but the input schema being null

src/Microsoft.Azure.Functions.Extensions.Mcp/Utilities/McpInputSchemaJsonUtilities.cs
```diff
@@ -0,0 +1,130 @@
...
+    {
+        // Check if this is the default empty schema
+        if (!schema.TryGetProperty("properties", out var properties) ||
+            !schema.TryGetProperty("required", out var required))
+        {
+            return false;
+        }
+
+        // Default schema has empty properties and required arrays
+        return properties.ValueKind == JsonValueKind.Object &&
```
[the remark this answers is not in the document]
liliankasem: I don't think we should have this at all. We can instead just check if a worker schema is provided:

```
if (_toolAttribute.InputSchema == null)
{
    // No explicit schema - generate from properties
}
else
{
    // Use the provided schema - even if empty
}
```

src/Microsoft.Azure.Functions.Extensions.Mcp/Abstractions/IMcpTool.cs
liliankasem: Should add docs to these props e.g.

```chsarp
    /// <summary>
    /// Gets or sets whether the input schema should be provided by the worker instead of being generated from reflection.
    /// When set to true, the <see cref="InputSchema"/> property must be provided.
    /// When set to false (default), the extension generates the schema from the function parameters or <see cref="ToolProperties"/>.
    /// </summary>
    public bool UseWorkerInputSchema { get; set; } = false;

    /// <summary>
    /// Gets or sets the MCP tool input schema as a JSON element.
    /// This should only be set when <see cref="UseWorkerInputSchema"/> is true.
    /// The schema must be a valid JSON Schema object with type "object".
    /// When null and UseWorkerInputSchema is false, the schema is generated from function parameters.
    /// </summary>
    public JsonElement? InputSchema { get; set; }
```

src/Microsoft.Azure.Functions.Extensions.Mcp/Trigger/McpToolTriggerBinding.cs
```diff
@@ -132,16 +132,22 @@ private Transport GetTransportInformation(CallToolExecutionContext context)
...
+        var listener = new McpToolListener(context.Executor, context.Descriptor.ShortName, 
+            _toolAttribute.ToolName, _toolAttribute.Description, toolProperties, _toolAttribute.InputSchema ?? McpInputSchemaJsonUtilities.DefaultMcpToolSchema);
```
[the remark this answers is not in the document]
aishwaryabh: I think this is ok given that we map tool.Properties to an empty array[ here ](https://github.com/Azure/azure-functions-mcp-extension/blob/main/src/Microsoft.Azure.Functions.Extensions.Mcp/McpToolExtensions.cs#L14) if it's null

src/Microsoft.Azure.Functions.Extensions.Mcp/Trigger/McpToolTriggerBinding.cs
aishwaryabh: done

src/Microsoft.Azure.Functions.Extensions.Mcp/Abstractions/IMcpTool.cs
aishwaryabh: added to the `McpToolTriggerAttribute`

src/Microsoft.Azure.Functions.Extensions.Mcp/Trigger/McpToolTriggerBinding.cs
```diff
@@ -132,15 +133,52 @@ private Transport GetTransportInformation(CallToolExecutionContext context)
...
+            return null;
+        }
+        try
+        {
+            using var doc = JsonDocument.Parse(attribute.InputSchema);
+            return doc.RootElement.Clone();
+        }
+        catch
+        {
+            // If parsing fails, return null
```
liliankasem: Don't love that we're swallowing an exception here, will make it hard to figure out whats wrong. The error above "Input schema was not generated in the worker extension" could be incorrect when the worker did generate a schema but its broken for any reason.

```csharp
        catch (JsonException ex)
        {
            throw new InvalidOperationException(
                $"Failed to parse InputSchema for tool '{attribute.ToolName}'. Schema must be valid JSON.", ex);
        }
```

src/Microsoft.Azure.Functions.Extensions.Mcp/Trigger/McpToolListener.cs
```diff
@@ -26,6 +27,23 @@ internal sealed class McpToolListener(ITriggeredFunctionExecutor executor,
...
     public ICollection<IMcpToolProperty> Properties { get; set; } = properties;
 
+    public JsonElement? InputSchema 
+    { 
+        get => _inputSchema; 
+        set
+        {
+            if (value.HasValue && !McpInputSchemaJsonUtilities.IsValidMcpToolSchema(value.Value))
+            {
+                throw new ArgumentException(
```
liliankasem: I know I had suggested something like this prior but I think its better we validate this schema where we read it and make this read only `public JsonElement? InputSchema => _inputSchema;`

So validation would actually happen on "GetInputSchema"

```csharp
        // Validate that the parsed schema is a valid MCP tool input schema
        if (!McpInputSchemaJsonUtilities.IsValidMcpToolSchema(schema))
        {
            throw new ArgumentException(
                "The specified document is not a valid MCP tool input JSON schema.",
                nameof(attribute.InputSchema));
        }

        return schema;
```

liliankasem (commented):
Looking good, I think one more pass here and I am ready to approve.

@jviau mind reviewing as well if you have time?

src/Microsoft.Azure.Functions.Extensions.Mcp/Trigger/McpToolTriggerBinding.cs
```diff
@@ -132,15 +133,52 @@ private Transport GetTransportInformation(CallToolExecutionContext context)
...
+
+        var listener = new McpToolListener(context.Executor, context.Descriptor.ShortName,
+            _toolAttribute.ToolName, _toolAttribute.Description, toolProperties, inputSchema);
 
         _toolRegistry.Register(listener);
 
         return Task.FromResult<IListener>(listener);
     }
 
+    private static JsonElement? GetInputSchema(McpToolTriggerAttribute attribute)
```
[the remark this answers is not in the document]
aishwaryabh: done!

jviau (commented):
There is an opportunity to refactor this PR a bit. Instead of passing the schema JsonElement and property collection around, with both the listener and the binding needing to know what both of them represent, we can abstract that away. To me it looks like the schema `JsonElement` and that property collection are exclusive from one another, you only use one or the other.

1. Create abstract class `ToolRequestValidator` with a `Validate` method on it (and appropriate params)
2. Have one implementation be `DefaultToolRequestValidator`, which uses the property collection.
3. Another implementation be `JsonSchemaToolRequestValidator`, which uses the JSON schema
4. The listener can now accept just a `ToolRequestValidator`, with the binding supplying the correct one based on the attribute

Rename the above classes as you see fit. But the general idea is to abstract away the flow of config -> validation method selection -> performing validation on request

jviau:
Json schema does more than validate required properties. It can also constrain the contents of a property as well. Are there any plans for supporting that?

src/Microsoft.Azure.Functions.Extensions.Mcp/Trigger/McpToolTriggerBinding.cs
```diff
@@ -132,15 +134,62 @@ private Transport GetTransportInformation(CallToolExecutionContext context)
...
+    internal static JsonElement? GetInputSchema(McpToolTriggerAttribute attribute)
+    {
+        if (string.IsNullOrEmpty(attribute.InputSchema))
+        {
+            return null;
+        }
+        try
+        {
+            using var doc = JsonDocument.Parse(attribute.InputSchema);
+            var schema =  doc.RootElement.Clone();
```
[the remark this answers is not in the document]
aishwaryabh: So I don't think we can pass `JsonDocument` from the worker. Initially I wanted to pass `JsonElement` directly from the worker to webjobs but the value would not persist when going through the host so we had to pass it in as a string and rehydrate it

aishwaryabh:
> Json schema does more than validate required properties. It can also constrain the contents of a property as well. Are there any plans for supporting that?

Yes there's plans to support this in the future but not in this PR!

src/Microsoft.Azure.Functions.Extensions.Mcp/Validation/PropertyBasedToolRequestValidator.cs
```diff
@@ -0,0 +1,36 @@
...
+        _properties = properties ?? throw new ArgumentNullException(nameof(properties));
+    }
+
+    /// <summary>
+    /// Gets the list of required property names for validation.
+    /// </summary>
+    /// <returns>A collection of required property names.</returns>
+    protected override IReadOnlyCollection<string> GetRequiredProperties()
+    {
+        return _properties
```
jviau: Same here, should this be calculated once (probably on construction) and then cached?

jviau (changes requested):
Make sure to flow `IDisposable` so that the `JsonDocument` is disposed at the correct time. It is also important that we switched to holding the `JsonDocument`, because the root element also holds onto the document. The previous approach of disposing the document but then returning the root element could have led into issues.

src/Microsoft.Azure.Functions.Extensions.Mcp/Trigger/McpToolListener.cs
```diff
@@ -24,7 +25,11 @@ internal sealed class McpToolListener(ITriggeredFunctionExecutor executor,
 
     public string? Description { get; set; } = toolDescription;
 
-    public ICollection<IMcpToolProperty> Properties { get; set; } = properties;
+    public ICollection<IMcpToolProperty> Properties { get; set; } = [];
```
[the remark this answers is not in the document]
aishwaryabh: `InputSchema` and `Properties` are being used in `McpToolExtensions` [here](https://github.com/Azure/azure-functions-mcp-extension/pull/136/changes#diff-ded6435da699c581f1bceb609238c4c83f86116cd5a9d65a8e5119ca52f6f366R15) to return the input schema whenever we need to generate the input schema, which is done when we we list the MCP tools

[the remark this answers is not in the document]
jviau: Ah. We can rename the validator class structure you added to a `ToolInputSchema` and have it expose both `Validate(...)` as well as `public abstract JsonElement Schema { get; }`

src/Microsoft.Azure.Functions.Extensions.Mcp/Validation/JsonSchemaToolRequestValidator.cs
```diff
@@ -0,0 +1,33 @@
...
+
+using System.Text.Json;
+using ModelContextProtocol.Protocol;
+
+namespace Microsoft.Azure.Functions.Extensions.Mcp.Validation;
+
+/// <summary>
+/// Validates tool request arguments using a JSON schema.
+/// </summary>
+internal sealed class JsonSchemaToolRequestValidator : ToolRequestValidator
```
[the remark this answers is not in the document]
aishwaryabh: done!

src/Microsoft.Azure.Functions.Extensions.Mcp/ToolInputSchema/PropertyBasedToolInputSchema.cs
```diff
@@ -0,0 +1,129 @@
...
+{
+    private readonly ICollection<IMcpToolProperty> _properties;
+
+    /// <summary>
+    /// Initializes a new instance of the PropertyBasedToolInputSchema class.
+    /// </summary>
+    /// <param name="properties">The tool properties to use for validation.</param>
+    public PropertyBasedToolInputSchema(ICollection<IMcpToolProperty> properties)
+    {
+        _properties = properties ?? throw new ArgumentNullException(nameof(properties));
```
jviau: consider taking in `IEnumerable` and then `ToList`'ing it here. This way you accept any enumerable, and then make a copy.

jviau (dismissed):

src/Microsoft.Azure.Functions.Extensions.Mcp/ToolInputSchema/PropertyBasedToolInputSchema.cs
aishwaryabh: done

src/Microsoft.Azure.Functions.Extensions.Mcp/ToolInputSchema/PropertyBasedToolInputSchema.cs
```diff
@@ -11,13 +10,86 @@ namespace Microsoft.Azure.Functions.Extensions.Mcp.Validation;
...
+            }
+
+            writer.WriteEndArray();
+
+            writer.WriteEndObject();
+        }
+
+        ms.Position = 0;
+        using var doc = JsonDocument.Parse(ms);
+        return doc.RootElement.Clone();
```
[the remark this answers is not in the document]
aishwaryabh: discussed offline and this is only available for .NET 10

jviau (approved):

PR #164, opened 2025-12-11
Implement resource registry

### LINKED ISSUE #156: Implement resource registry implementation
- Create IMcpResource interface
- Create IResourceRegistry interface 
- Implement DefaultResourceRegistry with security URI validation

liliankasem:

resolves #156

### Pull request checklist

### Additional information

Implements a resource registry system for MCP resources.

- IMcpResource: Core interface representing an MCP resource with URI, name, MIME type, description, size, and metadata collection
- IMcpResourceMetadata: Interface defining metadata items with Key and Value properties
- IResourceRegistry: Registry interface for managing resource registration, lookup, and listing operations

### DIFF
--- a/src/Microsoft.Azure.Functions.Extensions.Mcp/DefaultResourceRegistry.cs
+++ b/src/Microsoft.Azure.Functions.Extensions.Mcp/DefaultResourceRegistry.cs
@@ -39,7 +39,7 @@ internal sealed class DefaultResourceRegistry : IResourceRegistry
     }
 
     /// <inheritdoc/>
-    public ICollection<IMcpResource> GetResources()
+    public IReadOnlyCollection<IMcpResource> GetResources()
     {
         return _resources.Values;
     }


src/Microsoft.Azure.Functions.Extensions.Mcp/Abstractions/IMcpResourceMetadata.cs
```diff
@@ -0,0 +1,20 @@
+// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License.
+
+namespace Microsoft.Azure.Functions.Extensions.Mcp;
+
+/// <summary>
+/// Represents metadata associated with an MCP resource.
+/// </summary>
+public interface IMcpResourceMetadata
```
liliankasem: Created this for use by a future `McpResourceMetadata` attribute

src/Microsoft.Azure.Functions.Extensions.Mcp/Abstractions/IMcpResource.cs
```diff
@@ -0,0 +1,51 @@
...
+    /// </summary>
+    public IReadOnlyCollection<IMcpResourceMetadata>? Metadata { get; set; }
+
+    /// <summary>
+    /// Handles a read request for the resource.
+    /// </summary>
+    /// <param name="readResourceRequest">The read resource request context.</param>
+    /// <param name="cancellationToken">A cancellation token.</param>
+    /// <returns>A task that represents the asynchronous read operation. The task result contains the read
+    Task<ReadResourceResult> RunAsync(RequestContext<ReadResourceRequestParams> readResourceRequest, CancellationToken cancellationToken);
```
liliankasem: Thoughts on `ReadAsync` instead?

aishwaryabh: I prefer `ReadAsync` over `RunAsync`!

src/Microsoft.Azure.Functions.Extensions.Mcp/Abstractions/IMcpResource.cs
```diff
@@ -0,0 +1,51 @@
...
+
+/// <summary>
+/// Represents a Model Context Protocol (MCP) resource.
+/// </summary>
+internal interface IMcpResource
+{
+    /// <summary>
+    /// Gets the unique URI of the resource.
+    /// </summary>
+    string Uri { get; }
```
aishwaryabh: is there a reason we don't have identifiers for all of these properties? i know they're implied but was just curious since we're usually explicit with them in core-tools and other repos

liliankasem: Do you mean access modifier?  Yeah so all members of an interface are implicitly public - we can be explicit but I think its cleaner not to include something redundant - could go either way with this one but leaning to leave as is

src/Microsoft.Azure.Functions.Extensions.Mcp/Abstractions/IMcpResource.cs
```diff
@@ -0,0 +1,51 @@
...
+
+using ModelContextProtocol.Protocol;
+using ModelContextProtocol.Server;
+
+namespace Microsoft.Azure.Functions.Extensions.Mcp;
+
+/// <summary>
+/// Represents a Model Context Protocol (MCP) resource.
+/// </summary>
+internal interface IMcpResource
```
jviau: A lot of the properties on this interface are mutable. Do we want that? I prefer immutability unless it absolutely needs to be mutable.

liliankasem: Yup fair, I think at the moment I don't think they are mutable - perhaps only metadata depending on how we are going to handle that in later PRs. Happy to make them immutability to start and we can modify as needed.

src/Microsoft.Azure.Functions.Extensions.Mcp/Abstractions/IResourceRegistry.cs
```diff
@@ -0,0 +1,37 @@
...
+    /// </summary>
+    /// <param name="uri">The URI of the resource.</param>
+    /// <param name="resource">The retrieved resource, if found.</param>
+    /// <returns>True if the resource was found; otherwise, false.</returns>
+    bool TryGetResource(string uri, [NotNullWhen(true)] out IMcpResource? resource);
+
+    /// <summary>
+    /// Gets all registered MCP resources.
+    /// </summary>
+    ICollection<IMcpResource> GetResources();
```
jviau: Should probably be a read-only type, so no caller expects to mutate it.

```suggestion
    IReadOnlyCollection<IMcpResource> GetResources();
```

Also, what is the difference between `GetResources` and `ListResourcesAsync`?

jviau (approved):

src/Microsoft.Azure.Functions.Extensions.Mcp/Abstractions/IResourceRegistry.cs
jviau: Having one be async and the other not gives a bit of weird ergonomics. They are directly related (as your implementation shows, one is a projection of another). But the one being non-async effectively makes the other one non-async as well in real use scenarios.

PR #172, opened 2025-12-18
Structured content support

### LINKED ISSUE #151: Implement support for structured content
#114 and #113 added support for unstructured return types. We should also support structured content return types.

https://modelcontextprotocol.io/specification/2025-06-18/server/tools#structured-content

aishwaryabh:

resolves #151
Creates structured along with unstructured content
image

### Pull request checklist

### Additional information

The StructuredContent property should be added to CallToolResult during runtime invocation within the worker SDK. We can check the result type after invoking the function and check to see if the customer has added the attribute `McpResult` that is not CallToolResult or inheriting from ContentBlock. If the check passes, we then populate StructuredContent.

### DIFF
--- a/src/Microsoft.Azure.Functions.Worker.Extensions.Mcp.Sdk/FunctionsMiddleware/FunctionsMcpToolResultMiddleware.cs
+++ b/src/Microsoft.Azure.Functions.Worker.Extensions.Mcp.Sdk/FunctionsMiddleware/FunctionsMcpToolResultMiddleware.cs
@@ -68,6 +68,7 @@ internal class FunctionsMcpToolResultMiddleware : IFunctionsWorkerMiddleware
                 // Don't process CallToolResult - just serialize as-is
                 type = Constants.CallToolResultType;
                 content = JsonSerializer.Serialize(callToolResult, McpJsonUtilities.DefaultOptions);
+                structuredContent = callToolResult.StructuredContent?.ToJsonString();
                 break;
 
             case ContentBlock block:
--- a/src/release_notes.md
+++ b/src/release_notes.md
@@ -10,5 +10,8 @@
 
 ### Microsoft.Azure.Functions.Worker.Extensions.Mcp <version>
 
+- Added structured content support for tools (#172)
 
 ### Microsoft.Azure.Functions.Worker.Extensions.Mcp.Sdk <version>
+
+- Added structured content support for tools (#172)
--- a/test/Worker.Extensions.Mcp.Sdk.Tests/FunctionsMcpToolResultMiddlewareTests.cs
+++ b/test/Worker.Extensions.Mcp.Sdk.Tests/FunctionsMcpToolResultMiddlewareTests.cs
@@ -776,7 +776,7 @@ public class FunctionsMcpToolResultMiddlewareTests
         McpToolResult? mcpToolResult = JsonSerializer.Deserialize(result, McpJsonContext.Default.McpToolResult);
         Assert.NotNull(mcpToolResult);
         Assert.Equal(Constants.CallToolResultType, mcpToolResult.Type);
-        Assert.Null(mcpToolResult.StructuredContent); // Not preserved in McpToolResult wrapper
+        Assert.NotNull(mcpToolResult.StructuredContent); // StructuredContent is propagated to wrapper
 
         // Verify the CallToolResult is serialized in content
         var deserializedCallToolResult = JsonSerializer.Deserialize<CallToolResult>(mcpToolResult.Content!, McpJsonUtilities.DefaultOptions);
@@ -817,6 +817,7 @@ public class FunctionsMcpToolResultMiddlewareTests
         McpToolResult? mcpToolResult = JsonSerializer.Deserialize(result, McpJsonContext.Default.McpToolResult);
         Assert.NotNull(mcpToolResult);
         Assert.Equal(Constants.CallToolResultType, mcpToolResult.Type);
+        Assert.NotNull(mcpToolResult.StructuredContent); // StructuredContent is propagated to wrapper
 
         // Deserialize the CallToolResult from content
         var deserializedCallToolResult = JsonSerializer.Deserialize<CallToolResult>(mcpToolResult.Content!, McpJsonUtilities.DefaultOptions);
@@ -861,6 +862,49 @@ public class FunctionsMcpToolResultMiddlewareTests
         Assert.NotNull(deserializedCallToolResult);
         Assert.Equal(2, deserializedCallToolResult.Content.Count);
         Assert.Null(deserializedCallToolResult.StructuredContent);
+    }
+
+    [Fact]
+    public async Task Invoke_WithCallToolResultStructuredContent_PropagatesStructuredContentToWrapper()
+    {
+        // Arrange
+        var context = CreateMcpFunctionContext();
+        var structuredData = new { Name = "Test", Value = 123 };
+        var structuredJson = JsonSerializer.Serialize(structuredData);
+
+        var callToolResult = new CallToolResult
+        {
+            Content = new List<ContentBlock>
+            {
+                new TextContentBlock { Text = structuredJson }
+            },
+            StructuredContent = System.Text.Json.Nodes.JsonNode.Parse(structuredJson)
+        };
+
+        // Act
+        await _middleware.Invoke(context, ctx =>
+        {
+            SetInvocationResult(ctx, callToolResult);
+            return Task.CompletedTask;
+        });
+
+        // Assert
+        var result = _currentResult as string;
+        Assert.NotNull(result);
+
+        McpToolResult? mcpToolResult = JsonSerializer.Deserialize(result, McpJsonContext.Default.McpToolResult);
+        Assert.NotNull(mcpToolResult);
+        Assert.Equal(Constants.CallToolResultType, mcpToolResult.Type);
+
+        // Verify the wrapper's StructuredContent is populated from CallToolResult
+        Assert.NotNull(mcpToolResult.StructuredContent);
+        Assert.Contains("Test", mcpToolResult.StructuredContent);
+        Assert.Contains("123", mcpToolResult.StructuredContent);
+
+        // Also verify the inner CallToolResult still has StructuredContent
+        var deserializedCallToolResult = JsonSerializer.Deserialize<CallToolResult>(mcpToolResult.Content!, McpJsonUtilities.DefaultOptions);
+        Assert.NotNull(deserializedCallToolResult);
+        Assert.NotNull(deserializedCallToolResult.StructuredContent);
     }
 
     [Fact]
@@ -1504,6 +1548,7 @@ public class FunctionsMcpToolResultMiddlewareTests
             McpToolResult? mcpToolResult = JsonSerializer.Deserialize(result, McpJsonContext.Default.McpToolResult);
             Assert.NotNull(mcpToolResult);
             Assert.Equal(Constants.CallToolResultType, mcpToolResult.Type);
+            Assert.NotNull(mcpToolResult.StructuredContent); // StructuredContent is propagated to wrapper
 
             // Verify the CallToolResult is serialized correctly
             var deserializedCallToolResult = JsonSerializer.Deserialize<CallToolResult>(mcpToolResult.Content!, McpJsonUtilities.DefaultOptions);


src/Microsoft.Azure.Functions.Worker.Extensions.Mcp.Sdk/FunctionsMiddleware/FunctionsMcpToolResultMiddleware.cs
```diff
@@ -45,21 +45,115 @@ public async Task Invoke(FunctionContext context, FunctionExecutionDelegate next
...
+                if (callToolResult.StructuredContent != null)
                 {
-                    Text = functionResult is string s ? s : JsonSerializer.Serialize(functionResult)
-                }, McpJsonUtilities.DefaultOptions))
-        };
+                    structuredContent = callToolResult.StructuredContent.ToJsonString();
+                    
+                    // Ensure there's a TextContent block for backwards compatibility
+                    var hasTextContent = callToolResult.Content?.Any(c => c is TextContentBlock) ?? false;
+                    if (!hasTextContent)
```
aishwaryabh: do we need to do this validation if the user has provided `CallToolResult`?

src/Microsoft.Azure.Functions.Extensions.Mcp/Extensions.Mcp.csproj
```diff
@@ -17,7 +17,7 @@
     <PackageReference Include="Microsoft.Azure.WebJobs.Script.Abstractions" Version="1.0.4-preview" />
     <PackageReference Include="Microsoft.Bcl.AsyncInterfaces" Version="10.0.0" />
     <PackageReference Include="Microsoft.Extensions.Azure" Version="1.13.0" />
-    <PackageReference Include="ModelContextProtocol" Version="0.4.0-preview.3" />
+    <PackageReference Include="ModelContextProtocol" Version="0.5.0-preview.1" />
```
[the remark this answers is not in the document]
aishwaryabh: yeah I was experimenting before with the new version so will change this back!

src/Microsoft.Azure.Functions.Worker.Extensions.Mcp.Sdk/FunctionsMiddleware/FunctionsMcpToolResultMiddleware.cs
```diff
@@ -38,21 +38,77 @@ public async Task Invoke(FunctionContext context, FunctionExecutionDelegate next
...
+                content = JsonSerializer.Serialize(block, McpJsonUtilities.DefaultOptions);
+                break;
+
+            case IList<ContentBlock> blocks:
+                type = Constants.MultiContentResult;
+                content = JsonSerializer.Serialize(blocks, McpJsonUtilities.DefaultOptions);
+                break;
+
+            default:
+                // For other types, check if they have structured content properties
```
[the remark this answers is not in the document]
aishwaryabh: fair but this is the return type object so if the user wants to return both Contentblock and StructuredContent they can just directly return callToolResult

src/Microsoft.Azure.Functions.Extensions.Mcp/Trigger/ToolReturnValueBinder.cs
```diff
@@ -73,4 +76,16 @@ private static IList<ContentBlock> DeserializeToContentBlockCollection(McpToolRe
...
         return [contentBlock];
     }
+
+    private static JsonNode? DeserializeToStructuredContent(McpToolResult result)
+    {
+        JsonNode? node = null;
+        if (!string.IsNullOrEmpty(result.StructuredContent))
+        {
+            node = JsonSerializer.Deserialize<JsonNode>(result.StructuredContent!, McpJsonUtilities.DefaultOptions);
+
```
liliankasem: nit: empty line

src/Microsoft.Azure.Functions.Worker.Extensions.Mcp.Sdk/FunctionsMiddleware/FunctionsMcpToolResultMiddleware.cs
```diff
@@ -45,21 +45,115 @@ public async Task Invoke(FunctionContext context, FunctionExecutionDelegate next
...
+        string? content;
+        string? structuredContent = null;
+
+        // Determine type, content, and structured content based on the function result
+        switch (functionResult)
         {
-            ContentBlock block => (block.Type, JsonSerializer.Serialize(block, McpJsonUtilities.DefaultOptions)),
-            IList<ContentBlock> blocks => (Constants.MultiContentResult, JsonSerializer.Serialize(blocks, McpJsonUtilities.DefaultOptions)),
-            _ => (Constants.TextContextResult, JsonSerializer.Serialize(new TextContentBlock
+            case CallToolResult callToolResult:
```
liliankasem: I don't think we support a customer returning a CallToolResult directly today - is the host ext updated to support this?

src/Microsoft.Azure.Functions.Worker.Extensions.Mcp.Sdk/FunctionsMiddleware/FunctionsMcpToolResultMiddleware.cs
liliankasem: Not sure I understand the question? Should we validate the mcp spec?

src/Microsoft.Azure.Functions.Worker.Extensions.Mcp.Sdk/FunctionsMiddleware/FunctionsMcpToolResultMiddleware.cs
```diff
@@ -45,21 +45,115 @@ public async Task Invoke(FunctionContext context, FunctionExecutionDelegate next
...
+        {
+            Type = type,
+            Content = content,
+            StructuredContent = structuredContent
+        };
 
         _resultAccessor.SetResult(context, JsonSerializer.Serialize(mcpToolResult, McpJsonContext.Default.McpToolResult));
     }
 
+    private static bool IsPoco(object obj)
```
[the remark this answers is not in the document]
liliankasem: I feel like @jviau always has good ideas for things like this

src/Microsoft.Azure.Functions.Worker.Extensions.Mcp.Sdk/FunctionsMiddleware/FunctionsMcpToolResultMiddleware.cs
```diff
@@ -45,21 +45,115 @@ public async Task Invoke(FunctionContext context, FunctionExecutionDelegate next
...
-        };
+                    structuredContent = callToolResult.StructuredContent.ToJsonString();
+                    
+                    // Ensure there's a TextContent block for backwards compatibility
+                    var hasTextContent = callToolResult.Content?.Any(c => c is TextContentBlock) ?? false;
+                    if (!hasTextContent)
+                    {
+                        throw new InvalidOperationException(
+                            "CallToolResult contains StructuredContent but no TextContent block. " +
+                            "Per MCP specification, a tool that returns structured content SHOULD also " +
```
liliankasem: I don't think we need to reference the mcp spec

aishwaryabh: removed

src/Microsoft.Azure.Functions.Worker.Extensions.Mcp.Sdk/FunctionsMiddleware/FunctionsMcpToolResultMiddleware.cs
```diff
@@ -45,21 +46,123 @@ public async Task Invoke(FunctionContext context, FunctionExecutionDelegate next
...
+        var type = Constants.TextContextResult;
+        var content = JsonSerializer.Serialize(new TextContentBlock
+            {
+                Text = text
+            }, McpJsonUtilities.DefaultOptions);
+
+        return (type, content, structuredContent);
+    }
+
+    private static bool ShouldCreateStructuredContent(object obj)
```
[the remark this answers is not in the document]
aishwaryabh: Removed the inherited attribution and dictionary use-cases!

src/Microsoft.Azure.Functions.Worker.Extensions.Mcp.Sdk/FunctionsMiddleware/FunctionsMcpToolResultMiddleware.cs
```diff
@@ -45,21 +46,123 @@ public async Task Invoke(FunctionContext context, FunctionExecutionDelegate next
...
         {
-            ContentBlock block => (block.Type, JsonSerializer.Serialize(block, McpJsonUtilities.DefaultOptions)),
-            IList<ContentBlock> blocks => (Constants.MultiContentResult, JsonSerializer.Serialize(blocks, McpJsonUtilities.DefaultOptions)),
-            _ => (Constants.TextContextResult, JsonSerializer.Serialize(new TextContentBlock
-                {
-                    Text = functionResult is string s ? s : JsonSerializer.Serialize(functionResult)
-                }, McpJsonUtilities.DefaultOptions))
-        };
+            case CallToolResult callToolResult:
+                // Don't process CallToolResult - just serialize as-is
```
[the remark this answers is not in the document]
aishwaryabh: Checked and you're correct so I swapped the check to be in the worker sdk!

src/Microsoft.Azure.Functions.Worker.Extensions.Mcp.Sdk/FunctionsMiddleware/FunctionsMcpToolResultMiddleware.cs
```diff
@@ -45,21 +46,210 @@ public async Task Invoke(FunctionContext context, FunctionExecutionDelegate next
...
+        string? structuredContent = null;
+
+        // Determine type, content, and structured content based on the function result
+        switch (functionResult)
         {
-            ContentBlock block => (block.Type, JsonSerializer.Serialize(block, McpJsonUtilities.DefaultOptions)),
-            IList<ContentBlock> blocks => (Constants.MultiContentResult, JsonSerializer.Serialize(blocks, McpJsonUtilities.DefaultOptions)),
-            _ => (Constants.TextContextResult, JsonSerializer.Serialize(new TextContentBlock
+            case CallToolResult callToolResult:
+                // Validate if structured content exists, ensure text content exists for backwards compatibility
```
liliankasem: Backwards compatability with what?

src/Microsoft.Azure.Functions.Worker.Extensions.Mcp.Sdk/FunctionsMiddleware/FunctionsMcpToolResultMiddleware.cs
```diff
@@ -45,21 +46,210 @@ public async Task Invoke(FunctionContext context, FunctionExecutionDelegate next
...
+    /// </list>
+    /// </remarks>
+    /// <param name="obj">The object to evaluate.</param>
+    /// <returns>True if structured content should be created; otherwise, false.</returns>
+    private static bool ShouldCreateStructuredContent(object obj)
+    {
+        var type = obj.GetType();
+        return HasMcpResultAttributeRecursive(type);
+    }
+
```
liliankasem: Can we move all these helper methods out of the middleware? We can even have a type extension class

src/Microsoft.Azure.Functions.Worker.Extensions.Mcp/McpResultAttribute.cs
```diff
@@ -0,0 +1,18 @@
...
+/// Marks a POCO type as an MCP result type that should be serialized as structured content.
+/// When a function returns an object of a type decorated with this attribute,
+/// the result will be serialized as both text content (for backwards compatibility)
+/// and structured content (for clients that support it).
+/// </summary>
+/// <remarks>
+/// This attribute can be applied to classes, structs, and record types.
+/// </remarks>
+[AttributeUsage(AttributeTargets.Class | AttributeTargets.Struct, AllowMultiple = false)]
+public sealed class McpResultAttribute : Attribute
```
liliankasem: What's going on with this attribute? Is it specifically for structred content? If so, McpResult feels to vague

src/Microsoft.Azure.Functions.Worker.Extensions.Mcp.Sdk/FunctionsMiddleware/FunctionsMcpToolResultMiddleware.cs
aishwaryabh: This check is to satisfy this check in the MCP spec:
> For backwards compatibility, a tool that returns structured content SHOULD also return the serialized JSON in a TextContent block.

If the user returns `CallToolResult`, we don't want to modify their object or add text content to the passed in object, so that's why we have a check here.

src/Microsoft.Azure.Functions.Worker.Extensions.Mcp/McpResultAttribute.cs
liliankasem: I think I am still confused, the fluent API makes sense, but what has that got to do with the attribute? Do you mean that they go hand in hand with each other (but not required) i.e. you'd use the API to define the schema, then provide the content via [McpResult]. Does it make sense for this to then be a StructuredContent attribute?

src/Microsoft.Azure.Functions.Worker.Extensions.Mcp.Sdk/FunctionsMiddleware/FunctionsMcpToolResultMiddleware.cs
```diff
@@ -45,28 +45,151 @@ public async Task Invoke(FunctionContext context, FunctionExecutionDelegate next
...
+        if (ShouldCreateStructuredContent(functionResult))
+        {
+            // If there is a McpContentAttribute, create structured content
+            text = JsonSerializer.Serialize(functionResult, McpJsonUtilities.DefaultOptions);
+            structuredContent = text;
+        }
+        else
+        {
+            // For primitives and strings, convert to text
+            text = functionResult is string s ? s : JsonSerializer.Serialize(functionResult, McpJsonUtilities.DefaultOptions);
```
jviau: This special handling of `string` when everything else is JSON has caused issues in the past. I question if its the right approach.

What does the deserialization side look like? Does it also JSON deserialize? Is it deserialized? The issues we have ran into before is the receiving side always tries to JSON deserialize, which works in all cases, until it is a `string`, then it fails because it isn't a double-quoted string that JSON expects.

src/Microsoft.Azure.Functions.Worker.Extensions.Mcp.Sdk/StructuredContentTypeExtensions.cs
```diff
@@ -0,0 +1,96 @@
...
+    internal static bool HasMcpContentAttributeRecursive(this Type type)
+    {
+        // Check if the type itself is decorated with McpContentAttribute (no inheritance)
+        if (type.HasMcpContentAttribute())
+        {
+            return true;
+        }
+
+        // Skip dictionary types entirely
+        if (type.IsDictionaryType())
```
jviau: If we have collection support (which I am on the fence of), we should have dictionary support.

I think my last comment which led to you removing this was potential a misunderstanding. I thought you were checking for just `KeyValuePair<,>` and not `IEnumerable<KeyValuePair<,>>`. The later seems fine to me. The former doesn't.

jviau (approved):

PR #188, opened 2026-01-28
Refactor CI to break out worker release

liliankasem:
Now that we are going to be releasing worker extension in preview, we need to break out the versioning from the host ext version. I also broke out the CI release pipeline so that we can release the worker alone without a host ext.

### DIFF
--- a/src/Microsoft.Azure.Functions.Worker.Extensions.Mcp/Worker.Extensions.Mcp.csproj
+++ b/src/Microsoft.Azure.Functions.Worker.Extensions.Mcp/Worker.Extensions.Mcp.csproj
@@ -12,10 +12,11 @@
   </ItemGroup>
 
   <Target Name="IncludeExtensionInformationAttribute" BeforeTargets="GenerateAdditionalSources">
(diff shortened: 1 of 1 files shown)


eng/ci/templates/jobs/release-packages-host.yml
```diff
@@ -0,0 +1,69 @@
...
+    - stage: Prepare
+      jobs:
+        - job: Prepare
+          templateContext:
+            type: validationJob
+
+          steps:
+          - checkout: self
+          - pwsh: |
+              [xml]$versionProps = Get-Content "$(Build.SourcesDirectory)/src/Microsoft.Azure.Functions.Extensions.Mcp/Directory.Version.props"
```
jviau: Parsing `Directory.Version.props` is potentially inaccurate, as other targets can influence the final value. Since this is a release pipeline only, I recommend parsing the version from the nupkg being released.

Here is how I did it in dotnet-worker: https://github.com/Azure/azure-functions-dotnet-worker/blob/main/eng/ci/official-release.yml#L113

fabiocav (approved):

PR #166, opened 2025-12-12
Implement worker side input schema generation

aishwaryabh:
### Issue describing the changes in this PR

resolves #167

Note that the base branch will be `main` once #136 is merged in

### Pull request checklist

* [x] My changes **do not** require documentation changes
  * [ ] Otherwise: Documentation issue linked to PR
* [ ] My changes **should not** be added to the release notes for the next release
  * [x] Otherwise: I've added my notes to `release_notes.md`
* [x] I have added all required tests (Unit tests, E2E tests)

### Additional information

Additional PR information

PR #203, opened 2026-02-24
Output Schema Support

### LINKED ISSUE #96: Ability to specify outputSchema
### Description

aishwaryabh:

resolves #96 

Note base branch will change to `main` once PR #166 is in

### Pull request checklist

### Additional information

Additional PR information


src/Microsoft.Azure.Functions.Extensions.Mcp/ToolInputSchema/ToolOutputSchemaValidator.cs
```diff
@@ -0,0 +1,181 @@
...
+                $"Structured content is missing required properties declared in the output schema. Please provide: {names}",
+                McpErrorCode.InvalidParams);
+        }
+    }
+
+    /// <summary>
+    /// Validates that the types of properties in the structured content match
+    /// the types declared in the output schema's "properties" object.
+    /// </summary>
+    private static void ValidatePropertyTypes(JsonElement outputSchema, JsonObject structuredContent)
```
aishwaryabh: I added our own native support for validating structured content but lmk if we want to use another library instead. I thought that was overkill since our check is pretty simple for just making sure required types are present and the type matches

src/Microsoft.Azure.Functions.Worker.Extensions.Mcp/McpOutputAttribute.cs
```diff
@@ -0,0 +1,27 @@
...
+/// to clients during <c>tools/list</c>.
+/// </para>
+/// <para>
+/// This attribute can be applied to classes, structs, and record types.
+/// An explicit <c>WithOutputSchema</c> configuration via the fluent API takes precedence
+/// over the auto-generated schema.
+/// </para>
+/// </remarks>
+[AttributeUsage(AttributeTargets.Class | AttributeTargets.Struct, AllowMultiple = false, Inherited = true)]
+public sealed class McpOutputAttribute : McpContentAttribute
```
aishwaryabh: also created this attribute which inherits from `McpContent` that denotes that the class it decorates is supposed to be output schema AND structured content. Let me know if this is the correct approach

src/Microsoft.Azure.Functions.Extensions.Mcp/ToolInputSchema/ToolOutputSchemaValidator.cs
```diff
@@ -0,0 +1,181 @@
...
+                JsonValueKind.Number => "number",
+                JsonValueKind.True or JsonValueKind.False => "boolean",
+                JsonValueKind.Null => "null",
+                _ => "unknown"
+            },
+            _ => "unknown"
+        };
+    }
+
+    private static bool IsNullNode(JsonNode? node)
```
liliankasem: nit: this is only used in exactly one place (ValidateRequiredProperties) and is just a null check - I don't think this should be its own method

src/Microsoft.Azure.Functions.Extensions.Mcp/ToolInputSchema/ToolOutputSchemaValidator.cs
liliankasem: I think Fabio's general preference is not to use 3rd party libs if it can be avoided, but I do like the idea of using something like [JsonSchema.Net](https://www.nuget.org/packages/JsonSchema.Net) to do this validation. This could be as simple as

```csharp
var schema = JsonSchema.FromText(attribute.OutputSchema);
var result = schema.Evaluate(structuredContent);
if (!result.IsValid) { /* throw */ }
```

This way we can ensure complete spec coverage as I think the current approach has gaps (no $ref, no anyOf etc.). Do we expect users to supply simple schemas (flat objects, basic types), or potentially complex ones with $ref, nested objects, anyOf, etc.?

PR #208, opened 2026-03-17
File Resource Abstraction

### LINKED ISSUE #158: Implement File resource abstraction
Similar to the base `TextResourceContents` and `BlobResourceContents` types, we want to introduce a `FileResourceContents` type that abstracts away file I/O operations. This type would allow users to specify a file path, and the framework would handle the file loading and conversion automatically.

For example:

```csharp
return new FileResourceContents
{
    Uri = "ui://widget/welcome.html",
    MimeType = "text/html+skybridge",
    Path = <provde file path you want to serve up for this resource>
    Meta =
    {
        ["openai/widgetPrefersBorder"] = true
    }
};
```

When a function returns FileResourceContents, the framework would:

1. Read the file from the specified `Path`
2. Determine the appropriate content type based on `MimeType`
3. Convert it to either:
  - `TextResourceContents` for text-based MIME types (text/html, text/plain, etc.)
  - `BlobResourceContents` for binary files (images, PDFs, etc.), with automatic base64 encoding
4. Preserve the `Uri`, `MimeType`, and `Meta` properties in the final result

Alternatively we can ask the user to specifiy blob or text.

aishwaryabh:

resolves #158

### Pull request checklist

### Additional information
Taken from README:
When a function returns `FileResourceContents`, the framework:

- reads the file at `Path`
- uses `MimeType` to decide whether the content should be treated as text or binary
- converts text-like MIME types to `TextResourceContents`
- converts binary MIME types to `BlobResourceContents` with base64-encoded payloads
- preserves `Uri`, `MimeType`, and `Meta` on the final MCP response

The `Meta` object is emitted on the resulting MCP resource content as `_meta`, including nested JSON objects.

If `MimeType` is omitted on `FileResourceContents`, the extension falls back to the trigger attribute MIME type and then to file-extension-based content type detection.


aishwaryabh:
Closing PR to revisit design of original issue

PR #212, opened 2026-03-23
 Add MCP prompt SDK support for rich return types

### LINKED ISSUE #235: Add MCP prompt SDK support for rich return types
## Problem

Currently, prompt functions in the isolated worker model can only return plain strings. The host-side `PromptReturnValueBinder` wraps any string return into a single `PromptMessage` with `Role.User` and `TextContent`. This means developers cannot:

1. **Return multiple messages** — prompts that need a system message followed by a user message (e.g., a code review prompt with instructions + context) must be flattened into a single string.
2. **Set message roles** — all prompt responses are forced to `Role.User`. There is no way to return `Role.Assistant` messages for few-shot examples or system instructions.
3. **Return `GetPromptResult` with a description** — the prompt description from `[McpPromptTrigger]` is static. There is no way to dynamically set or override the description based on arguments.
4. **Return structured content** — `ImageContent`, `EmbeddedResource`, or other content types cannot be included in prompt responses.

This is inconsistent with tool functions, which already support rich return types (`CallToolResult`, `ContentBlock`, `ImageContentBlock`, etc.) via `FunctionsMcpToolResultMiddleware` in the SDK package.

## Proposed Solution

Add `FunctionsMcpPromptResultMiddleware` to the `Worker.Extensions.Mcp.Sdk` package, following the same pattern as the existing tool result middleware:

- Intercept prompt function results in the worker pipeline
- Serialize them into an `McpPromptResult` wrapper format
- Host-side `PromptReturnValueBinder` deserializes and processes the wrapper

### Supported return types

| Return Type | Behavior |
|-------------|----------|
| `GetPromptResult` | Serialized directly with description and messages |
| `PromptMessage` | Wrapped in a single-element message list |
| `IList<PromptMessage>` | Serialized as message array |
| `string` / `object` | Wrapped as text in a `Role.User` `PromptMessage` (default, backwards compatible) |

## Acceptance Criteria

- [ ] `FunctionsMcpPromptResultMiddleware` added to SDK package
- [ ] Host-side `PromptReturnValueBinder` handles `McpPromptResult` wrapper deserialization
- [ ] All four return types work end-to-end
- [ ] Backwards compatible — existing string-returning prompts continue to work
- [ ] Unit and E2E tests added

liliankasem:
> ⚠️ **MCP version skew — non-text prompt content does not round-trip**
 >
 > The host extension still depends on `ModelContextProtocol` **0.4.0-preview.3**, while this SDK pulls in 
**1.2.0**. The two versions encode `byte`-backed content blocks (`ImageContentBlock.Data`, `AudioContentBlock.Data`, `BlobResourceContents.Blob`) differently on the wire — 1.2.0 produces a base64 string, 0.4.0 a JSON byte array.
 >
 > A worker function that returns `GetPromptResult` / `IList<PromptMessage>` containing **non-text** content blocks 
will serialize fine via the SDK middleware, decode fine in the host, but the host then re-serializes those blocks to
 the MCP client with its older types, producing JSON the client rejects (`"Invalid Base64 string"`).
 >
 > **For this PR, only text content is supported end-to-end.** Image / audio / embedded-resource content in prompt 
results requires aligning the host on `ModelContextProtocol` 1.2.0 — tracked separately.

Resolves #235

### Pull request checklist

### Additional information

Adds `FunctionsMcpPromptResultMiddleware` to the `Worker.Extensions.Mcp.Sdk` package, enabling prompt functions to return rich types (`GetPromptResult`, `PromptMessage`, `IList<PromptMessage>`) instead of only plain strings.

The middleware follows the same pattern as `FunctionsMcpToolResultMiddleware`:
- Intercepts prompt function results in the worker pipeline
- Serializes them into `McpPromptResult` wrapper format
- Host-side `PromptReturnValueBinder` deserializes and processes the wrapper

#### Supported return types

| Return Type | Behavior |
|-------------|----------|
| `GetPromptResult` | Serialized directly with description and messages |
| `PromptMessage` | Wrapped in a single-element message list |
| `IList<PromptMessage>` | Serialized as message array |
| `string` / `object` | Wrapped as text in a `Role.User` `PromptMessage` (default, backwards compatible) |

### DIFF
--- a/src/Microsoft.Azure.Functions.Extensions.Mcp/Trigger/McpPromptResult.cs
+++ b/src/Microsoft.Azure.Functions.Extensions.Mcp/Trigger/McpPromptResult.cs
@@ -1,0 +1,23 @@
+// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License.
+
+namespace Microsoft.Azure.Functions.Extensions.Mcp;
+
+/// <summary>
+/// Host-side counterpart of the worker SDK's <c>McpPromptResult</c> envelope.
+/// Carries a serialized payload (e.g. a <c>GetPromptResult</c> or list of
+/// <c>PromptMessage</c>) along with a <see cref="Type"/> discriminator describing
+/// how to deserialize <see cref="Content"/>.
+/// </summary>
+internal sealed class McpPromptResult
+{
+    /// <summary>
+    /// The serialized inner payload.
+    /// </summary>
+    public string? Content { get; set; }
+
+    /// <summary>
+    /// Discriminator describing the shape of <see cref="Content"/>.
+    /// </summary>
+    public required string Type { get; init; }
+}
--- a/src/Microsoft.Azure.Functions.Worker.Extensions.Mcp.Sdk/FunctionsMiddleware/FunctionsMcpPromptResultMiddleware.cs
+++ b/src/Microsoft.Azure.Functions.Worker.Extensions.Mcp.Sdk/FunctionsMiddleware/FunctionsMcpPromptResultMiddleware.cs
@@ -10,14 +10,13 @@ using ModelContextProtocol.Protocol;
 namespace Microsoft.Azure.Functions.Worker.Extensions.Mcp;
 
 /// <summary>
-/// Normalizes the worker's prompt-function return value into a serialized
-/// <see cref="GetPromptResult"/> JSON string for the host binder to consume.
+/// Wraps the worker's prompt-function return value into an <see cref="McpPromptResult"/>
+/// envelope (Type + serialized Content) for the host binder to deserialize.
 /// </summary>
 /// <remarks>
-/// Unlike <see cref="FunctionsMcpToolResultMiddleware"/>, no envelope is needed:
-/// every supported prompt return type collapses without information loss into
-/// the MCP protocol's own <see cref="GetPromptResult"/> shape, which is also
-/// the natural cross-language wire contract.
+/// Mirrors <see cref="FunctionsMcpToolResultMiddleware"/>. The envelope's <c>Type</c>
+/// discriminator lets the host accept any successfully deserialized payload — including
+/// empty <c>GetPromptResult</c> values — without shape-sniffing the JSON.
 /// </remarks>
 internal class FunctionsMcpPromptResultMiddleware : IFunctionsWorkerMiddleware
 {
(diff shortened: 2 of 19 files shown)


src/Microsoft.Azure.Functions.Extensions.Mcp/Trigger/McpPromptResult.cs
```diff
@@ -0,0 +1,21 @@
...
+/// <summary>
+/// Represents the result returned by an MCP prompt execution.
+/// Utilized when rich content type support is enabled via the SDK extension.
+/// </summary>
+public sealed class McpPromptResult
+{
+    /// <summary>
+    /// The serialized content returned by the prompt.
+    /// </summary>
+    public string? Content { get; set; }
```
satvu, line 15: This differs from the other `McpPromptResult` class (no `[JsonPropertyName]`). Is that intentional?

src/Microsoft.Azure.Functions.Extensions.Mcp/Trigger/McpPromptListener.cs
```diff
@@ -59,6 +59,12 @@ public async Task<GetPromptResult> GetAsync(RequestContext<GetPromptRequestParam
         }
 
         var promptResult = await execution.ResultTask;
-        return promptResult;
+
+        if (promptResult is GetPromptResult getPromptResult)
```
satvu, line 63: In what cases is the result returned not of this type? What scenario is this?

liliankasem: It should always be GetPromptResult, but this is more of a defensive safety check

src/Microsoft.Azure.Functions.Extensions.Mcp/Trigger/McpPromptResult.cs
liliankasem: Yes, it's intentional and follows the same pattern as the existing McpToolResult classes.

1. we need to recreate it because they're (host ext and the .Sdk ext) completely separate packages with no reference between them
2.  the Sdk version needs explicit [JsonPropertyName] attributes because it controls the wire format. It serializes the result that gets sent across the worker → host boundary. The host-side version doesn't need them because it deserializes using McpJsonSerializerOptions.DefaultOptions

satvu (dismissed):

src/Microsoft.Azure.Functions.Worker.Extensions.Mcp.Sdk/FunctionsMiddleware/FunctionsMcpPromptResultMiddleware.cs
```diff
@@ -0,0 +1,105 @@
...
+            return;
+        }
+
+        string type;
+        string? content;
+
+        switch (functionResult)
+        {
+            case GetPromptResult getPromptResult:
+                type = Constants.GetPromptResultType;
```
aishwaryabh, line 60: nit: this is the same as the constants in the host. Perhaps we should have a centralized shared package for ensuring that they're the same? I don't think this is a pressing change for now so we can consider this non-blocking and open a folowup PR

src/Microsoft.Azure.Functions.Extensions.Mcp/Trigger/McpPromptResult.cs
```diff
@@ -0,0 +1,21 @@
+// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License.
+
+namespace Microsoft.Azure.Functions.Extensions.Mcp;
+
+/// <summary>
+/// Represents the result returned by an MCP prompt execution.
+/// Utilized when rich content type support is enabled via the SDK extension.
+/// </summary>
+public sealed class McpPromptResult
```
aishwaryabh: ai flagged this for me that this should technically be internal

test/Worker.Extensions.Mcp.Sdk.Tests/FunctionsMcpPromptResultMiddlewareTests.cs
aishwaryabh, line 1: AI flagged a test coverage gap for PromptMessage.Content since it appears we only test with `TextContentBlock`. Can we add test for image/resource content as well?

src/Microsoft.Azure.Functions.Worker.Extensions.Mcp.Sdk/FunctionsMiddleware/FunctionsMcpPromptResultMiddleware.cs
liliankasem: Agree the duplication isn't ideal since these strings form a wire contract between the two packages. That said, a shared package is not the right way to go for something like this. We should file an issue and figure out another solution for this.

aishwaryabh (dismissed):

test/Extensions.Mcp.Tests/PromptReturnValueBinderTests.cs
```diff
@@ -107,6 +104,21 @@ public async Task SetValueAsync_WithJsonWithoutMessagesField_DeserializesAsEmpty
...
+    {
+        var executionContext = GetPromptExecutionContextHelper.CreateExecutionContext();
+        var binder = new PromptReturnValueBinder(executionContext);
+
+        await binder.SetValueAsync("", CancellationToken.None);
+
+        var result = await executionContext.ResultTask;
+        Assert.NotNull(result);
+        Assert.Single(result.Messages);
+        var content = Assert.IsType<TextContentBlock>(result.Messages[0].Content);
```
aishwaryabh: question: do prompts only support text content blocks or do we wanna account for other types like image/audio/blob?

src/Microsoft.Azure.Functions.Worker.Extensions.Mcp.Sdk/FunctionsMiddleware/FunctionsMcpPromptResultMiddleware.cs
```diff
@@ -0,0 +1,92 @@
+// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License.
+
```
aishwaryabh, line 3: AI flagged the following use case for me:

The worker now serializes prompt return values as structured JSON.
A valid empty result (messages is an empty array) is possible and should remain structured.
The host currently only accepts deserialized results when there is at least one message or a description.
Because of that rule, a valid empty structured result is treated as plain text.
The raw JSON is then wrapped as a user message instead of being preserved as an empty structured prompt result.
Why This Matters
It silently changes intended behavior.
Clients expecting an empty structured response get a text message containing JSON.
This can cause confusing downstream behavior and contract mismatches.
Recommended Fix
Treat any successfully deserialized prompt result as valid, including empty ones.

=====================
Is accepting an empty prompt a valid use case?

aishwaryabh (approved):

PR #241, opened 2026-04-20
Output Schema Support

### LINKED ISSUE #96: Ability to specify outputSchema

aishwaryabh:

resolves #96 

### Pull request checklist

### Additional information

Additional PR information


src/Microsoft.Azure.Functions.Worker.Extensions.Mcp.Sdk/StructuredContentTypeExtensions.cs
```diff
@@ -18,4 +18,14 @@ internal static bool HasMcpContentAttribute(this Type type)
         // Only check for direct attribution, not inherited
         return type.GetCustomAttributes(typeof(McpContentAttribute), inherit: false).Length > 0;
     }
+
+    /// <summary>
+    /// Checks if the type is decorated with <see cref="McpOutputAttribute"/>.
+    /// </summary>
+    /// <param name="type">The type to check.</param>
+    /// <returns>True if the type has the attribute; otherwise, false.</returns>
+    internal static bool HasMcpOutputAttribute(this Type type)
```
liliankasem, line 27: This isn't called anywhere?

src/Microsoft.Azure.Functions.Worker.Extensions.Mcp/McpContentAttribute.cs
```diff
@@ -11,8 +11,9 @@ namespace Microsoft.Azure.Functions.Worker.Extensions.Mcp;
 /// </summary>
 /// <remarks>
 /// This attribute can be applied to classes, structs, and record types.
+/// It supports inheritance: a derived type will inherit the attribute from its base type.
 /// </remarks>
-[AttributeUsage(AttributeTargets.Class | AttributeTargets.Struct, AllowMultiple = false)]
-public sealed class McpContentAttribute : Attribute
+[AttributeUsage(AttributeTargets.Class | AttributeTargets.Struct, AllowMultiple = false, Inherited = true)]
+public class McpContentAttribute : Attribute
```
liliankasem, line 17: What's with the change in this file? What are you trying to do?

src/Microsoft.Azure.Functions.Worker.Extensions.Mcp/McpOutputAttribute.cs
```diff
@@ -0,0 +1,19 @@
...
+/// When a function's return type (or its inner type for <see cref="System.Threading.Tasks.Task{T}"/>)
+/// is decorated with this attribute, the MCP extension will auto-generate a JSON Schema
+/// from the type and include it as the tool's output schema.
+/// </summary>
+/// <remarks>
+/// This attribute can be applied to classes, structs, and record types.
+/// The generated schema will have root <c>"type": "object"</c> and include all public properties.
+/// </remarks>
+[AttributeUsage(AttributeTargets.Class | AttributeTargets.Struct, AllowMultiple = false, Inherited = true)]
+public sealed class McpOutputAttribute : Attribute
```
liliankasem, line 17: Flagged by agent:

Inconsistent `inherit:` flags — `HasMcpOutputAttribute` and `OutputSchemaGenerator` use `inherit:true`; `HasMcpContentAttribute` uses `inherit:false`

src/Microsoft.Azure.Functions.Worker.Extensions.Mcp/Configuration/OutputSchemaGenerator.cs
```diff
@@ -0,0 +1,102 @@
...
+
+            if (returnType is null || !HasMcpOutputAttribute(returnType))
+            {
+                return false;
+            }
+
+            outputSchema = GenerateSchemaFromType(returnType);
+            return outputSchema is not null;
+        }
+        catch (Exception ex)
```
liliankasem, line 51: Do we need to catch all exceptions? What could go wrong here that needs to be addressed i.e. `JsonException/InvalidOperationException` we dont want to swallow real bugs

test/Worker.Extensions.Mcp.Tests/Configuration/ResolveToolOutputSchemaExtensionTests.cs
```diff
@@ -0,0 +1,83 @@
+// Copyright (c) Microsoft Corporation. All rights reserved.
```
liliankasem, line 1: Missing test cases:


1. POCO with `[JsonPropertyName]` / `[JsonIgnore]` / `[JsonRequired]` — verify the generated schema matches what STJ actually does.
1. Nullable refs (`string?` vs `string`) — pin the mapping.
1. Enum (default int) vs `[JsonStringEnumConverter]` — assert the generator and runtime agree.
1. Self-referential POCO — make sure we don't `StackOverflow` and that `$ref` is emitted properly.
1. `Task`, `void`, primitive returns — no schema, and pin `UnwrapReturnType`'s behavior.
1. `WithOutputSchema<T>()` where `T` is an interface / abstract / `string` — pin behavior.
1. `class Derived : BaseWithMcpOutput {}` — verify the derived class gets a schema (once the `inherit:` story is consistent).
1. `WithOutputSchema(...)` + `[McpOutput]` return type on the same tool — pin which one wins (explicit, I'd assume).

src/Microsoft.Azure.Functions.Extensions.Mcp/DefaultToolRegistry.cs
```diff
@@ -41,6 +41,7 @@ public ValueTask<ListToolsResult> ListToolsAsync(CancellationToken cancellationT
                 Name = tool.Name,
                 Description = tool.Description,
                 InputSchema = tool.GetPropertiesInputSchema(),
+                OutputSchema = tool.GetOutputSchemaElement(),
```
liliankasem, line 44: If a tool sets OutputSchema but ends up with UseResultSchema=false (e.g. it has output bindings, so the SDK transformer leaves the flag
 off), the host will advertise an outputSchema on tools/list but every tools/call returns text-only content with no structuredContent. That violates the 2025-06-18 spec guidance.

src/Microsoft.Azure.Functions.Extensions.Mcp/Abstractions/IMcpTool.cs
```diff
@@ -24,6 +25,11 @@ internal interface IMcpTool
     /// </summary>
     public ToolInputSchema ToolInputSchema { get; }
 
+    /// <summary>
+    /// Gets the output schema for the tool result, if any.
+    /// </summary>
+    public JsonDocument? OutputSchema { get; }
```
liliankasem, line 31: Why not JsonElement?

src/Microsoft.Azure.Functions.Extensions.Mcp/Trigger/McpToolTriggerBinding.cs
```diff
@@ -180,6 +182,16 @@ private ToolInputSchema CreateToolInputSchema()
         }
     }
 
+    internal static JsonDocument? GetOutputSchema(McpToolTriggerAttribute attribute)
```
liliankasem, line 185: We should just use JsonElement here

```csharp
     using var doc = JsonDocument.Parse(attribute.OutputSchema);
     return doc.RootElement.Clone();
```

src/Microsoft.Azure.Functions.Extensions.Mcp/Trigger/McpToolTriggerAttribute.cs
```diff
@@ -40,9 +40,16 @@ public sealed class McpToolTriggerAttribute(string toolName, string? description
...
+    /// Gets or sets the MCP tool output schema as a JSON string.
+    /// This should only be set when the worker provides an output schema.
+    /// The schema JSON string must be a valid JSON Schema.
+    /// </summary>
+    public string? OutputSchema { get; set; }
+
     /// <summary>
     /// Gets or sets whether the result schema should be provided by the worker instead of being generated by the host extension.
-    /// When set to true, the <see cref="ResultSchema"/> property must be provided.
+    /// When set to true, the <see cref="OutputSchema"/> property must be provided.
```
liliankasem, line 52: Why is this comment changed? `UseResultSchema` is for `ResultSchema` - it is not tied to output schema?

src/Microsoft.Azure.Functions.Extensions.Mcp/Trigger/McpToolTriggerBinding.cs
```diff
@@ -180,6 +182,16 @@ private ToolInputSchema CreateToolInputSchema()
...
+    {
+        if (string.IsNullOrEmpty(attribute.OutputSchema))
+        {
+            return null;
+        }
+
+        return JsonDocument.Parse(attribute.OutputSchema);
```
liliankasem, line 192: We should have a try/catch here;  a malformed schema string surfaces as a raw JsonException from deep inside CreateListenerAsync, with no tool context making it hard to debug. We can handle very similar to input schema. e.g.

```csharp
// input schema code
 catch (JsonException ex)
 {
     doc?.Dispose();
     throw new InvalidOperationException(
         $"Failed to parse InputSchema for tool '{attribute.ToolName}'. Schema must be valid JSON.", ex);
 }
```

src/Microsoft.Azure.Functions.Extensions.Mcp/Trigger/McpToolTriggerBinding.cs
liliankasem, line 185: We have no validation here. 

GetInputSchema calls `IsValidMcpToolSchema(doc)` to confirm
it's actually a JSON Schema (object root, etc.). GetOutputSchema accepts anything that parses. At minimum the root should be a JSON object, since a JSON Schema is always an object. Worth a parallel 
`IsValidMcpToolSchema` check (or a shared helper).

src/Microsoft.Azure.Functions.Extensions.Mcp/Trigger/McpToolTriggerBinding.cs
liliankasem: Also recommend having a finally to dispose

```
     finally
     {
         doc?.Dispose();
     }
```

This is a fix for the input schema side as well

src/Microsoft.Azure.Functions.Extensions.Mcp/McpToolExtensions.cs
```diff
@@ -12,4 +12,10 @@ internal static JsonElement GetPropertiesInputSchema(this IMcpTool tool)
         ArgumentNullException.ThrowIfNull(tool);
         return tool.ToolInputSchema.GetSchemaElement();
     }
+
+    internal static JsonElement? GetOutputSchemaElement(this IMcpTool tool)
```
liliankasem, line 16: With the change to JsonElement, either:

 - Delete GetOutputSchemaElement and have DefaultToolRegistry read tool.OutputSchema directly, or
 - Keep it for symmetry with the input helper (but it's just return tool.OutputSchema;)

src/Microsoft.Azure.Functions.Worker.Extensions.Mcp.Sdk/StructuredContentTypeExtensions.cs
liliankasem: the worker ext has its own copy

```
// OutputSchemaGenerator.cs
 private static bool HasMcpOutputAttribute(Type type)
     => type.GetCustomAttribute<McpOutputAttribute>(inherit: true) is not null;
```

So I don't know why we need this file at all to begin with? If `HasMcpContentAttribute` is used just once, we can just inline it ` type.GetCustomAttribute<McpContentAttribute>(inherit: false) is not null` so that we can clean up extra files

src/Microsoft.Azure.Functions.Worker.Extensions.Mcp/Configuration/McpBindingBuilder/Steps/ResolveToolOutputSchemaExtension.cs
```diff
@@ -0,0 +1,63 @@
...
+                        continue;
+                    }
+                }
+                catch (JsonException ex)
+                {
+                    context.Logger.LogWarning(ex,
+                        "Explicit output schema for tool '{ToolName}' is not valid JSON and will be ignored.",
+                        toolName);
+                }
+            }
```
liliankasem, line 51: A user who explicitly calls WithOutputSchema("garbage") gets a warning 
log, then silently falls through to reflection (or no schema at all). That's confusing, use of the  explicit `WithOutputSchema` call should probably hard-fail.

src/Microsoft.Azure.Functions.Worker.Extensions.Mcp/Configuration/McpBindingBuilder/Steps/ResolveToolOutputSchemaExtension.cs
```diff
@@ -0,0 +1,63 @@
...
+            var options = context.ToolOptions.Get(toolName);
+
+            // Priority 1: Explicit output schema (WithOutputSchema)
+            var explicitSchema = options.OutputSchema;
+            if (!string.IsNullOrWhiteSpace(explicitSchema))
+            {
+                try
+                {
+                    var parsed = JsonNode.Parse(explicitSchema);
+                    if (parsed is not null)
```
liliankasem, line 39: This just calls JsonNode.Parse and trusts the result. None of the four WithOutputSchema(...) overloads check that the root is type: "object", so today all of these are accepted by the worker:

- WithOutputSchema("\"hello\"") (primitive root)
- WithOutputSchema("[]") (array root)
- WithOutputSchema("{}") (no type at all)

I think we want an OutputSchemaValidator that mirrors InputSchemaValidator, called from all four overloads and from this pipeline step.

liliankasem, line 39: this check is only meaningful for the literal string "null" (which parses to JsonNode? = null)

src/Microsoft.Azure.Functions.Worker.Extensions.Mcp/Configuration/McpBindingBuilder/Steps/ResolveToolOutputSchemaExtension.cs
```diff
@@ -0,0 +1,63 @@
...
+                catch (JsonException ex)
+                {
+                    context.Logger.LogWarning(ex,
+                        "Explicit output schema for tool '{ToolName}' is not valid JSON and will be ignored.",
+                        toolName);
+                }
+            }
+
+            // Priority 2: Reflection-based (auto-generated from [McpOutput] on return type)
+            if (OutputSchemaGenerator.TryGenerateFromFunction(context.Function, out var outputSchema, context.Logger)
```
liliankasem, line 54: Similar to our convo about reflection in the input shcema, do we want to support that here? We can do a JsonSchemaExporter follow up in a second pr

src/Microsoft.Azure.Functions.Worker.Extensions.Mcp/Configuration/OutputSchemaGenerator.cs
```diff
@@ -0,0 +1,102 @@
...
+        return type.GetCustomAttribute<McpOutputAttribute>(inherit: true) is not null;
+    }
+
+    /// <summary>
+    /// Generates a JSON schema from the specified type using <see cref="JsonSchemaExporter"/>.
+    /// </summary>
+    internal static JsonNode? GenerateSchemaFromType(Type type)
+    {
+        var options = JsonSerializerOptions.Default;
+        var schemaNode = options.GetJsonSchemaAsNode(type, new JsonSchemaExporterOptions
```
liliankasem, line 95: The schema is generated from Default (PascalCase, no custom converters, no naming policy). But the worker actually serializes the response with whatever options the user/host has configured, typically camelCase, often with custom converters. Result: schema says { "MyProperty": ... }, payload contains { "myProperty": ... }, MCP clients reject the response as schema-violating. This bug fires for essentially every real workload because most Functions configs are non-default.

The method also takes no JsonSerializerOptions parameter, so even if a caller wanted to inject the right options, the API doesn't allow it. Both the signature and the body need fixing.

src/Microsoft.Azure.Functions.Worker.Extensions.Mcp/Configuration/OutputSchemaGenerator.cs
```diff
@@ -0,0 +1,102 @@
...
+
+        try
+        {
+            if (!FunctionMethodResolver.TryGetScriptRoot(out _))
+            {
+                return false;
+            }
+
+            if (!FunctionMethodResolver.TryResolveMethod(functionMetadata, out var method) || method is null)
+            {
```
liliankasem, line 37: When TryGetScriptRoot or TryResolveMethod returns false, the method returns false with 
no log - this could be hard for users to figure out so we should have a debug log

src/Microsoft.Azure.Functions.Worker.Extensions.Mcp/Configuration/OutputSchemaGenerator.cs
```diff
@@ -0,0 +1,102 @@
...
+
+    /// <summary>
+    /// Generates a JSON schema from the specified type using <see cref="JsonSchemaExporter"/>.
+    /// </summary>
+    internal static JsonNode? GenerateSchemaFromType(Type type)
+    {
+        var options = JsonSerializerOptions.Default;
+        var schemaNode = options.GetJsonSchemaAsNode(type, new JsonSchemaExporterOptions
+        {
+            TreatNullObliviousAsNonNullable = true,
```
liliankasem, line 97: AI flagged:

Wuestionable default. If a user's project doesn't have NRT enabled, every property becomes required in the schema, even logically optional ones. Schema says required, payload omits the 
field, validation fails. Either default to false or surface as an option.

src/Microsoft.Azure.Functions.Worker.Extensions.Mcp/Configuration/OutputSchemaGenerator.cs
```diff
@@ -0,0 +1,102 @@
...
+        }
+    }
+
+    /// <summary>
+    /// Unwraps the inner type from Task{T} or ValueTask{T}, returning the actual result type.
+    /// Returns null for void, Task, or ValueTask (no result type).
+    /// </summary>
+    internal static Type? UnwrapReturnType(Type returnType)
+    {
+        if (returnType == typeof(void) || returnType == typeof(Task) || returnType == typeof(ValueTask))
```
liliankasem, line 64: What about `IAsyncEnumerable<T>`?

src/Microsoft.Azure.Functions.Worker.Extensions.Mcp/DependencyInjection/McpToolBuilder.cs
```diff
@@ -108,4 +110,75 @@ public McpToolBuilder WithInputSchema(JsonNode schemaNode)
...
+    /// Sets an explicit JSON output schema for this tool from a <see cref="JsonNode"/>.
+    /// </summary>
+    /// <param name="schemaNode">A <see cref="JsonNode"/> representing a valid JSON schema.</param>
+    /// <returns>The current builder instance, enabling fluent configuration.</returns>
+    /// <exception cref="ArgumentNullException">Thrown when <paramref name="schemaNode"/> is null.</exception>
+    public McpToolBuilder WithOutputSchema(JsonNode schemaNode)
+    {
+        ArgumentNullException.ThrowIfNull(schemaNode, nameof(schemaNode));
+
+        var schemaJson = schemaNode.ToJsonString();
```
liliankasem, line 143: Need to have validation here

src/Microsoft.Azure.Functions.Worker.Extensions.Mcp/DependencyInjection/McpToolBuilder.cs
```diff
@@ -108,4 +110,75 @@ public McpToolBuilder WithInputSchema(JsonNode schemaNode)
...
+    }
+
+    /// <summary>
+    /// Generates and sets a JSON output schema from the specified CLR type using <see cref="JsonSchemaExporter"/>.
+    /// </summary>
+    /// <param name="type">The CLR type to generate the JSON schema from.</param>
+    /// <param name="serializerOptions">Optional <see cref="JsonSerializerOptions"/> to control schema generation.
+    /// When null, <see cref="JsonSerializerOptions.Default"/> is used.</param>
+    /// <returns>The current builder instance, enabling fluent configuration.</returns>
+    public McpToolBuilder WithOutputSchema(Type type, JsonSerializerOptions? serializerOptions = null)
```
liliankasem, line 156: Simialr to input schema, do we want to defer the type based path?

src/Microsoft.Azure.Functions.Worker.Extensions.Mcp/DependencyInjection/McpToolBuilder.cs
```diff
@@ -108,4 +110,75 @@ public McpToolBuilder WithInputSchema(JsonNode schemaNode)
...
+    {
+        ArgumentNullException.ThrowIfNull(type, nameof(type));
+
+        var options = serializerOptions ?? JsonSerializerOptions.Default;
```
liliankasem, line 160: `JsonSerializerOptions.Default` same comment as before about drift

src/Microsoft.Azure.Functions.Worker.Extensions.Mcp/DependencyInjection/McpToolBuilder.cs
```diff
@@ -108,4 +110,75 @@ public McpToolBuilder WithInputSchema(JsonNode schemaNode)
...
+        var schemaNode = options.GetJsonSchemaAsNode(type, new JsonSchemaExporterOptions
+        {
+            TreatNullObliviousAsNonNullable = true,
```
liliankasem, line 163: Users can't override without round-tripping through WithOutputSchema(JsonNode). Either expose JsonSchemaExporterOptions as a parameter, or default to false.

Also seeing duplicate code here so we should break out into a shared JsonSchemaExporterOptions

src/Microsoft.Azure.Functions.Worker.Extensions.Mcp/DependencyInjection/McpToolBuilder.cs
```diff
@@ -108,4 +110,75 @@ public McpToolBuilder WithInputSchema(JsonNode schemaNode)
...
         return this;
     }
+
+    /// <summary>
+    /// Sets an explicit JSON output schema for this tool.
+    /// The output schema describes the structure of the tool's result.
+    /// </summary>
+    /// <param name="jsonSchema">A valid JSON schema string defining the tool's output structure.</param>
+    /// <returns>The current builder instance, enabling fluent configuration.</returns>
+    /// <exception cref="ArgumentException">Thrown when the schema is null/empty or is not valid JSON.</exception>
```
liliankasem, line 120: Actuallly a JsonException too; ` JsonNode.Parse(jsonSchema)` can throw

src/Microsoft.Azure.Functions.Worker.Extensions.Mcp/DependencyInjection/McpToolBuilder.cs
```diff
@@ -108,4 +110,75 @@ public McpToolBuilder WithInputSchema(JsonNode schemaNode)
...
+    public McpToolBuilder WithOutputSchema(string jsonSchema)
+    {
+        ArgumentException.ThrowIfNullOrEmpty(jsonSchema, nameof(jsonSchema));
```
liliankasem, line 123: Doesnt catch whitespace

aishwaryabh:
Closing in favor of #245 and #246

PR #243, opened 2026-04-21
Add input schema support for the worker extension

### LINKED ISSUE #167: Worker changes for generating input schema for .NET isolated functions
If the `UseWorkerInputSchema` flag is enabled to true, generate the input schema for the native types in the worker.

aishwaryabh:
Lets users provide an explicit JSON input schema for an MCP tool via McpToolBuilder.WithInputSchema(string|JsonNode). When set, the worker emits the schema and useWorkerInputSchema=true on the tool trigger binding, which the host already understands via McpToolTriggerAttribute.

Schema validation mirrors host McpInputSchemaJsonUtilities.IsValidMcpToolSchema so a schema that the worker accepts will not be rejected by the host:
- root must be a JSON object
- type="object" required at root
- properties must be a JSON object when present
- required must be a JSON array when present

The new ResolveToolInputSchema pipeline step is wired after PatchPropertyBindings (no behavior change to the existing properties path). When no explicit schema is configured, the step is a no-op and the host continues to generate a schema from toolProperties.

Reflection-based / POCO-based auto schema generation is intentionally out of scope for this PR — it would risk drift between the worker schema and the actual host binder semantics.

Tools only. Prompts and resources don't yet have host-side input-schema support.

resolves #167 

### Pull request checklist

### Additional information

Additional PR information

### DIFF
--- a/src/Microsoft.Azure.Functions.Worker.Extensions.Mcp/Configuration/McpBindingBuilder/Steps/ResolveToolInputSchemaExtension.cs
+++ b/src/Microsoft.Azure.Functions.Worker.Extensions.Mcp/Configuration/McpBindingBuilder/Steps/ResolveToolInputSchemaExtension.cs
@@ -1,9 +1,6 @@
 // Copyright (c) Microsoft Corporation. All rights reserved.
 // Licensed under the MIT License.
 
-using System.Text.Json;
-using Microsoft.Extensions.Logging;
-
 namespace Microsoft.Azure.Functions.Worker.Extensions.Mcp.Configuration.Builders.Steps;
 
 /// <summary>
@@ -20,6 +17,8 @@ internal static class ResolveToolInputSchemaExtension
 {
     public static McpBindingBuilder ResolveToolInputSchema(this McpBindingBuilder builder)
     {
+        ArgumentNullException.ThrowIfNull(builder);
+
         var context = builder.Context;
 
         foreach (var binding in context.ToolTriggerBindings)
@@ -27,27 +26,12 @@ internal static class ResolveToolInputSchemaExtension
             var toolName = binding.Identifier!;
             var options = context.ToolOptions.Get(toolName);
 
-            var explicitSchema = options.InputSchema;
-            if (string.IsNullOrWhiteSpace(explicitSchema))
+            if (options.InputSchema is null)
             {
                 continue;
             }
 
-            // The schema was already validated by McpToolBuilder.WithInputSchema.
-            // Re-validate defensively in case ToolOptions was mutated directly.
-            try
-            {
-                _ = InputSchemaValidator.ValidateAndParse(explicitSchema, nameof(ToolOptions.InputSchema));
-            }
-            catch (Exception ex) when (ex is ArgumentException or JsonException)
-            {
-                context.Logger.LogWarning(ex,
-                    "Explicit input schema for tool '{ToolName}' is invalid and will be ignored.",
-                    toolName);
-                continue;
-            }
-
-            binding.InputSchema = explicitSchema;
+            binding.InputSchema = options.InputSchema.Json;
             binding.UseWorkerInputSchema = true;
         }
 
--- a/src/Microsoft.Azure.Functions.Worker.Extensions.Mcp/Configuration/McpBindingBuilder/Steps/ResolveToolInputSchemaExtension.cs
+++ b/src/Microsoft.Azure.Functions.Worker.Extensions.Mcp/Configuration/McpBindingBuilder/Steps/ResolveToolInputSchemaExtension.cs
@@ -2,7 +2,6 @@
 // Licensed under the MIT License.
 
 using System.Text.Json;
-using System.Text.Json.Nodes;
 using Microsoft.Extensions.Logging;
 
 namespace Microsoft.Azure.Functions.Worker.Extensions.Mcp.Configuration.Builders.Steps;
--- a/src/Microsoft.Azure.Functions.Worker.Extensions.Mcp/Configuration/McpBindingBuilder/Steps/ResolveToolInputSchemaExtension.cs
+++ b/src/Microsoft.Azure.Functions.Worker.Extensions.Mcp/Configuration/McpBindingBuilder/Steps/ResolveToolInputSchemaExtension.cs
@@ -1,0 +1,57 @@
+// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License.
+
+using System.Text.Json;
+using System.Text.Json.Nodes;
+using Microsoft.Extensions.Logging;
+
+namespace Microsoft.Azure.Functions.Worker.Extensions.Mcp.Configuration.Builders.Steps;
+
+/// <summary>
+/// For each <c>mcpToolTrigger</c> binding, applies an explicit input schema configured
+/// via <c>McpToolBuilder.WithInputSchema(...)</c> and signals the host to use it
+/// (<c>useWorkerInputSchema = true</c>).
+///
+/// <para>
+/// When no explicit schema is configured, this step is a no-op and the host continues
+/// to generate a schema from <c>toolProperties</c> as before.
+/// </para>
+/// </summary>
+internal static class ResolveToolInputSchemaExtension
+{
+    public static McpBindingBuilder ResolveToolInputSchema(this McpBindingBuilder builder)
+    {
+        var context = builder.Context;
+
+        foreach (var binding in context.ToolTriggerBindings)
+        {
+            var toolName = binding.Identifier!;
+            var options = context.ToolOptions.Get(toolName);
+
+            var explicitSchema = options.InputSchema;
+            if (string.IsNullOrWhiteSpace(explicitSchema))
+            {
+                continue;
+            }
+
+            // The schema was already validated by McpToolBuilder.WithInputSchema.
+            // Re-validate defensively in case ToolOptions was mutated directly.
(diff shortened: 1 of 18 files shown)


src/release_notes.md
```diff
@@ -16,6 +16,23 @@
 
 #### Changes
 
+- Add `WithInputSchema` fluent API for MCP tools (#243)
```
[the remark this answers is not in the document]
liliankasem: invalid, 243 is correct

liliankasem (dismissed):

src/Microsoft.Azure.Functions.Worker.Extensions.Mcp/Configuration/McpBindingBuilder/Steps/ResolveToolInputSchemaExtension.cs
```diff
@@ -0,0 +1,56 @@
...
+/// <para>
+/// When no explicit schema is configured, this step is a no-op and the host continues
+/// to generate a schema from <c>toolProperties</c> as before.
+/// </para>
+/// </summary>
+internal static class ResolveToolInputSchemaExtension
+{
+    public static McpBindingBuilder ResolveToolInputSchema(this McpBindingBuilder builder)
+    {
+        var context = builder.Context;
```
jviau, line 22: Consider adding param validation. `ArgumentNullException.ThrowIfNull(builder);`

src/Microsoft.Azure.Functions.Worker.Extensions.Mcp/Configuration/McpBindingBuilder/Steps/ResolveToolInputSchemaExtension.cs
```diff
@@ -0,0 +1,56 @@
...
+            if (string.IsNullOrWhiteSpace(explicitSchema))
+            {
+                continue;
+            }
+
+            // The schema was already validated by McpToolBuilder.WithInputSchema.
+            // Re-validate defensively in case ToolOptions was mutated directly.
+            try
+            {
+                _ = InputSchemaValidator.ValidateAndParse(explicitSchema, nameof(ToolOptions.InputSchema));
```
jviau: Cases like this to me makes a case for using a strong type and not passing strings around. If you have an object in such a way that it can only be constructed IF the input is valid, then you can guarantee a validate-once and consume whenever approach.

An example of this problem is emails. Instead of having `string Email` on your user, and needing to validate its an OK email each time you go to use it, you can instead have a class `Email` and have `Email Email` on your type. The class `Email` can only be constructed IF the input string is valid. So now if you ever have a non-null `Email`, you know it is _already validated_ and safe to use. And then using it is simply accessing the (immutable) raw string backing the email.

src/Microsoft.Azure.Functions.Worker.Extensions.Mcp/DependencyInjection/McpToolBuilder.cs
```diff
@@ -62,4 +63,49 @@ public McpToolBuilder AsMcpApp(Action<IMcpAppBuilder> configure)
...
+    /// the explicit schema takes precedence on hosts that understand it, while older hosts
+    /// continue to consume <c>toolProperties</c>.
+    /// </remarks>
+    /// <param name="jsonSchema">A valid JSON schema string defining the tool's input parameters.</param>
+    /// <returns>The current <see cref="McpToolBuilder"/> instance, enabling fluent configuration.</returns>
+    /// <exception cref="ArgumentException">Thrown when the schema is null/empty or does not conform to MCP requirements.</exception>
+    /// <exception cref="System.Text.Json.JsonException">Thrown when the schema string is not valid JSON.</exception>
+    public McpToolBuilder WithInputSchema(string jsonSchema)
+    {
+        InputSchemaValidator.ValidateAndParse(jsonSchema, nameof(jsonSchema));
```
jviau: We could add validation to the setter of `InputSchema` directly, so there is no way around the validation and you can confidently use `InputSchema` later on knowing it is valid.

But also consider the strongly typed approach, it will help avoid any ambiguity of "is it valid or not" on its own (and not tied to the parent object having validation).

jviau (changes requested):
The validation approach can be done in a way to avoid needing to re-validate. For the quickest solution, I recommend adding the validation to the `ToolOptions.InputSchema` setter.

You can also consider using a strongly typed schema class which will perform validation on construction, but up to you on that.

src/Microsoft.Azure.Functions.Worker.Extensions.Mcp/DependencyInjection/McpToolBuilder.cs
liliankasem: I support going with strongly typed if it will clean things up. @aishwaryabh we would follow the same pattern for output schema in that case as well.

aishwaryabh: added strongly type validation!

jviau (dismissed):

src/release_notes.md
```diff
@@ -16,6 +16,23 @@
...
+
+    Lets users provide an explicit JSON input schema for an MCP tool via `McpToolBuilder.WithInputSchema(string|JsonNode)`. When set, the worker emits the schema and `useWorkerInputSchema=true` on the tool trigger binding, which the host uses instead of generating a schema from tool properties.
```
liliankasem:
```suggestion
    Lets users provide an explicit JSON input schema for an MCP tool via `McpToolBuilder.WithInputSchema(string|JsonNode)`.
```

src/release_notes.md
```diff
@@ -16,6 +16,23 @@
...
+
+    ```csharp
+    builder.ConfigureMcpTool("MyTool")
+        .WithInputSchema("""
```
liliankasem: Can we use a json node example instead?

src/Microsoft.Azure.Functions.Worker.Extensions.Mcp/Configuration/McpInputSchema.cs
```diff
@@ -0,0 +1,47 @@
...
+
+namespace Microsoft.Azure.Functions.Worker.Builder;
+
+/// <summary>
+/// A strongly-typed, validated MCP tool input JSON schema.
+/// An <see cref="McpInputSchema"/> instance can only exist when the schema
+/// conforms to the MCP tool input-schema shape expected by the host — once constructed,
+/// consumers can use <see cref="Json"/> without re-validating.
+/// </summary>
+public sealed class McpInputSchema
```
liliankasem, line 16: - No ToString() — for diagnostic logs (logger.Log("...{Schema}", schema)) you'll get 
McpInputSchema not the JSON. Override ToString() => Json;.
 - No equality — minor, but two ctors that both accept JsonNode.Parse(json) of the same 
string produce instances that aren't Equals. Probably never matters; add 
IEquatable<McpInputSchema> only if a test needs it.
 - No implicit conversion to string — also probably fine, but (string)schema would be a 
nice ergonomic. Skip if not asked for.

src/Microsoft.Azure.Functions.Worker.Extensions.Mcp/Configuration/ToolOptions.cs
```diff
@@ -44,4 +44,12 @@ public void AddProperty(string name, McpToolPropertyType type, string descriptio
...
     /// </summary>
     public AppOptions? AppOptions { get; set; }
+
+    /// <summary>
+    /// Gets or sets an explicit JSON input schema for the tool.
+    /// When set, the worker emits this schema (and <c>useWorkerInputSchema = true</c>)
+    /// on the tool trigger binding, taking precedence over any host-side schema
+    /// generated from <see cref="Properties"/>.
+    /// </summary>
+    public McpInputSchema? InputSchema { get; set; }
```
liliankasem, line 54: Can you validate this:

 If anything serializes ToolOptions to JSON (config snapshots, diagnostic logs, snapshot
 tests), this property's wire format changed when the type flipped from string? to 
 McpInputSchema? — it now serializes as { "Json": "..." } instead of a raw string. If that
 matters, a small JsonConverter on McpInputSchema that round-trips as the raw JSON string
 would preserve the old shape:
 
  [JsonConverter(typeof(McpInputSchemaJsonConverter))]
  public sealed class McpInputSchema { ... }
 
 If nothing downstream serializes ToolOptions, ignore.

--

the property will now serialize as:

 { "InputSchema": { "Json": "{\"type\":\"object\"...}" } }

…instead of:

 { "InputSchema": "{\"type\":\"object\"...}" }

liliankasem (changes requested):

src/Microsoft.Azure.Functions.Worker.Extensions.Mcp/Configuration/ToolOptions.cs
aishwaryabh: I don't think anything downstream serializes ToolOptions, so I think the converter may be overkill

src/Microsoft.Azure.Functions.Worker.Extensions.Mcp/Configuration/McpInputSchema.cs
aishwaryabh: addressed first one! I don't think the other two are relevant

liliankasem (approved):
lgtm - @jviau do you mind taking a second pass on this?

PR #244, opened 2026-04-22
Enforce required prompt arguments at protocol boundary

liliankasem:
McpPromptListener.GetAsync now validates required arguments declared on the prompt against the inbound GetPromptRequestParams.Arguments before invoking the function, mirroring the tool path's protocol-level validation in McpToolListener (ToolInputSchema.Validate).

Previously, required prompt arguments were only enforced via PromptArgumentInputBinding.BindAsync, which runs only when a worker parameter is decorated with [McpPromptArgument(IsRequired = true)]. That works for the .NET worker but provides no enforcement for TypeScript / Python / Java workers, which surface 'required' on prompts/list as advisory metadata only.

The thrown McpProtocolException uses InvalidParams to match the tool-side error semantics, and lists all missing required arguments in a single message.

### DIFF
--- a/src/Microsoft.Azure.Functions.Extensions.Mcp/Trigger/McpPromptListener.cs
+++ b/src/Microsoft.Azure.Functions.Extensions.Mcp/Trigger/McpPromptListener.cs
@@ -88,7 +88,7 @@ internal sealed class McpPromptListener(
 
             if (provided is null
                 || !provided.TryGetValue(declared.Name, out var value)
-                || IsValueNullOrUndefined(value))
+                || value.IsNullOrUndefined())
             {
                 (missing ??= []).Add(declared.Name);
             }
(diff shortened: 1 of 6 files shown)


src/Microsoft.Azure.Functions.Extensions.Mcp/Trigger/McpPromptListener.cs
```diff
@@ -61,4 +68,45 @@ public async Task<GetPromptResult> GetAsync(RequestContext<GetPromptRequestParam
...
+
+        if (missing is { Count: > 0 })
+        {
+            throw new McpProtocolException(
+                $"One or more required prompt arguments are missing values. Please provide: {string.Join(", ", missing)}",
+                McpErrorCode.InvalidParams);
+        }
+    }
+
+    private static bool IsValueNullOrUndefined(JsonElement value) => value.ValueKind switch
```
aishwaryabh: nit this is duplicated in ToolINputSchema I belive

aishwaryabh (dismissed):

src/Microsoft.Azure.Functions.Extensions.Mcp/Trigger/McpPromptListener.cs
liliankasem: let me consolidate these!

aishwaryabh (approved):

PR #245, opened 2026-04-22
Add output schema support on host extension

### LINKED ISSUE #96: Ability to specify outputSchema

aishwaryabh:
Exposes McpToolTriggerAttribute.OutputSchema (string) and plumbs it through McpToolTriggerBinding → McpToolListener → IMcpTool.OutputSchema (JsonElement?). Validates schema shape at binding time and wraps JsonException in InvalidOperationException with tool name context.

resolves #96 

### Pull request checklist

### Additional information

Additional PR information


src/release_notes.md
```diff
@@ -6,7 +6,7 @@
 
 ### Microsoft.Azure.Functions.Extensions <version>
 
-- <entry>
+- Added output schema support on the tool trigger. The `McpToolTriggerAttribute` now exposes an `OutputSchema` property, and `IMcpTool.OutputSchema` is surfaced as `JsonElement?` so the host can advertise a tool's structured result schema via the MCP protocol. Schema is validated at binding time and only applied when `UseResultSchema` is set. (#245)
```
liliankasem: Simplify this to `Added output schema support on the tool trigger. #245)`

test/Extensions.Mcp.Tests/McpToolTriggerBindingTests.cs
```diff
@@ -210,8 +210,10 @@ public void GetInputSchema_WithInvalidJson_ThrowsException()
         };
```
liliankasem, line 210: Missing new tests for all the new surface areas

src/Microsoft.Azure.Functions.Extensions.Mcp/Trigger/McpToolTriggerBinding.cs
```diff
@@ -114,14 +114,16 @@ private Transport GetTransportInformation(CallToolExecutionContext context) =>
     public Task<IListener> CreateListenerAsync(ListenerFactoryContext context)
     {
         ToolInputSchema inputSchema = CreateToolInputSchema();
+        JsonElement? outputSchema = _toolAttribute.UseResultSchema ? GetOutputSchema(_toolAttribute) : null;
```
liliankasem: We shouldn't be doing this.

UseResultSchema today is only set in the .NET worker and only if the customer is using the .Sdk package. This will make it so that we are rejecting outputschema for any other use case which means output schema can only work if you also use the result schema which should not be the case.

liliankasem: The original spec-compliance concern was: don't advertise outputSchema if the binder can't
produce structuredContent. That's a real problem, SimpleToolReturnValueBinder only emits
text. So the gate's intent is right, but UseResultSchema is the wrong gate. It conflates two orthogonal things:

 1. "Worker shapes the result envelope" (UseResultSchema's real meaning, the 
binder-selection flag).
 2. "Output is structured-content-capable" (what we actually want to gate on).

For .NET workers these happen to align (UseResultSchema=true ⇒ ToolReturnValueBinder ⇒ can
produce StructuredContent). For other workers they don't align at all, because their
result-envelope shaping happens differently.

liliankasem: A solution: Add host-side validation of outgoing structuredContent against the declared outputSchema (as we discussed for the eventual ToolOutputSchema wrapper). Then a misbehaving worker fails fast instead of confusing the client. Or, trust the worker to deliver structured content. Spec violation is on the worker, not the host.

src/Microsoft.Azure.Functions.Extensions.Mcp/Trigger/McpToolTriggerBinding.cs
```diff
@@ -158,26 +160,65 @@ private ToolInputSchema CreateToolInputSchema()
...
+            return null;
+        }
+
+        JsonDocument? doc = null;
+        try
+        {
+            doc = JsonDocument.Parse(attribute.OutputSchema);
+
+            // Validate that the parsed schema is a valid MCP tool schema (object root)
+            if (!McpInputSchemaJsonUtilities.IsValidMcpToolSchema(doc))
```
liliankasem: Given this is being reused between input and output we should rename this to `IsValidMcpToolJsonSchema`

src/Microsoft.Azure.Functions.Extensions.Mcp/Utilities/McpInputSchemaJsonUtilities.cs
```diff
@@ -11,11 +11,12 @@ namespace Microsoft.Azure.Functions.Extensions.Mcp;
 public static class McpInputSchemaJsonUtilities
```
liliankasem: Sorry when I suggested the name change, I didn't realize this was public, this should have been an internal class - it will be a break but we should make this internal now whilst its still early in the sdk. 

You can call it out as breaking as we did with the transformer:

https://github.com/Azure/azure-functions-mcp-extension/releases/tag/1.4.0

liliankasem (commented):
AI review comments:

1. Tests don't verify the Clone() lifecycle promise

The whole point of the Clone() in GetOutputSchema is that the returned JsonElement is
self-owning — no underlying JsonDocument to dispose. None of the 9 new tests touch a
returned element after the helper returns through GC pressure or assert it doesn't
reference an internal disposable. Probably fine in practice (the contract is implicit), but
a test like:

 [Fact]
 public void GetOutputSchema_ReturnedElement_IsSelfOwning_AfterGc()
 {
     var attribute = new McpToolTriggerAttribute("T", "D")
     {
         OutputSchema = """{ "type": "object", "properties": { "x": { "type": "string" } } 
}"""
     };
 
     var result = McpToolTriggerBinding.GetOutputSchema(attribute);
     GC.Collect();
     GC.WaitForPendingFinalizers();
     GC.Collect();
 
     // If the element still referenced an internal pooled buffer, this would throw or read
 garbage.
     Assert.Equal("object", result!.Value.GetProperty("type").GetString());
 }

Mostly defensive — the Clone() semantics are documented by STJ — but for a feature that
previously had a JsonDocument? OutputSchema lifecycle bug, one assertion that the new
contract holds is cheap insurance.

2. Empty-string test (line 707) — control character in the diff

The diff output shows [m OutputSchema = "" with a stray ANSI escape. Almost certainly a
diff-display artifact from terminal coloring, not the actual file. Just sanity-check the
file in an editor.

3. GetOutputSchemaElement extension method is still a pass-through

From the prior review — tool.OutputSchema and tool.GetOutputSchemaElement() return the
exact same value now. Not a bug, just clutter. Either delete (favorite) or keep for
symmetry with GetPropertiesInputSchema (defensible). Not blocking.

src/Microsoft.Azure.Functions.Extensions.Mcp/Utilities/McpInputSchemaJsonUtilities.cs
aishwaryabh: done!

src/Microsoft.Azure.Functions.Extensions.Mcp/Trigger/McpToolTriggerBinding.cs
aishwaryabh: Updated with option 2 to trust the worker to deliver structured content that is valid

test/Extensions.Mcp.Tests/McpToolTriggerBindingTests.cs
aishwaryabh: added!

src/Microsoft.Azure.Functions.Extensions.Mcp/Utilities/McpInputSchemaJsonUtilities.cs
```diff
@@ -8,14 +8,15 @@ namespace Microsoft.Azure.Functions.Extensions.Mcp;
 /// <summary>
 /// Utilities for working with MCP tool JSON schemas.
 /// </summary>
-public static class McpInputSchemaJsonUtilities
+internal static class McpInputSchemaJsonUtilities
```
kshyju, line 11: Breaking change?  Should we instead consider warning users of it's use instead?

```
[EditorBrowsable(EditorBrowsableState.Never)]
[Obsolete("Message about why this shouldn't be used.", error: false)]
```

src/Microsoft.Azure.Functions.Extensions.Mcp/Trigger/McpToolTriggerBinding.cs
```diff
@@ -158,26 +160,65 @@ private ToolInputSchema CreateToolInputSchema()
...
+            return doc.RootElement.Clone();
+        }
+        catch (JsonException ex)
+        {
+            throw new InvalidOperationException(
+                $"Failed to parse OutputSchema for tool '{attribute.ToolName}'. Schema must be valid JSON.", ex);
+        }
+        finally
+        {
+            doc?.Dispose();
```
kshyju: nit: you can ` using JsonDocument doc = JsonDocument.Parse(` and avoid explicit finally to dispose. (no functional difference)

kshyju (commented):
LGTM, except the breaking change. Should we discuss that with broader group?

src/Microsoft.Azure.Functions.Extensions.Mcp/Trigger/McpToolTriggerBinding.cs
```diff
@@ -158,26 +160,65 @@ private ToolInputSchema CreateToolInputSchema()
...
+    {
+        if (string.IsNullOrEmpty(attribute.OutputSchema))
+        {
+            return null;
+        }
+
+        JsonDocument? doc = null;
+        try
+        {
+            doc = JsonDocument.Parse(attribute.OutputSchema);
```
kshyju: Is `JsonElement.Parse` a better approach?

src/Microsoft.Azure.Functions.Extensions.Mcp/Utilities/McpInputSchemaJsonUtilities.cs
liliankasem: Yeah this was my suggestion: https://github.com/Azure/azure-functions-mcp-extension/pull/245#discussion_r3127071477

I had a similar conversation about another class in the worker ext with Fabio and we decided it was worth it to break in a minor early before the sdk has more adoption. Happy to have another conversation about this, but I do think we are fine here

src/Microsoft.Azure.Functions.Extensions.Mcp/Trigger/McpToolTriggerBinding.cs
aishwaryabh: This is what copilot said:

JsonElement.Parse doesn't exist as a static method — the closest equivalent is JsonSerializer.Deserialize<JsonElement>(...), which would give us a self-owning element directly. I considered it, but kept JsonDocument.Parse + RootElement.Clone() for a couple of reasons:

Symmetry with GetInputSchema — that method has to return a JsonDocument because it's stored long-lived in JsonSchemaToolInputSchema and disposed on listener teardown. Using the same parse pattern in both keeps the code parallel.
Shared validator — McpInputSchemaJsonUtilities.IsValidMcpToolJsonSchema takes a JsonDocument. Switching only GetOutputSchema to JsonElement would require either a second validator overload or changing the signature at both call sites and tests.
With the using + Clone() now in place, the returned JsonElement is already self-owning, so the readability gain from Deserialize<JsonElement> feels small relative to the churn.

src/Microsoft.Azure.Functions.Extensions.Mcp/Utilities/McpInputSchemaJsonUtilities.cs
aishwaryabh: agreed I think we're good here as long as we update release notes

src/Microsoft.Azure.Functions.Extensions.Mcp/Trigger/McpToolTriggerBinding.cs
```diff
@@ -158,26 +160,62 @@ private ToolInputSchema CreateToolInputSchema()
...
+        try
+        {
+            using JsonDocument doc = JsonDocument.Parse(attribute.OutputSchema);
+
+            // Validate that the parsed schema is a valid MCP tool schema (object root)
+            if (!McpInputSchemaJsonUtilities.IsValidMcpToolJsonSchema(doc))
+            {
+                throw new ArgumentException(
+                    "The specified document is not a valid MCP tool output JSON schema.",
+                    nameof(attribute.OutputSchema));
```
kshyju: Should this be `nameof(attribute)` ?

kshyju (dismissed):

src/release_notes.md
```diff
@@ -7,6 +7,9 @@
 ### Microsoft.Azure.Functions.Extensions <version>
 
 - Validate required prompt arguments (#244)
+- Added output schema support on the tool trigger. (#245)
+
+> ⚠️ **Breaking change**: `McpInputSchemaJsonUtilities` is now `internal` (was `public`). It was never intended as a public API. Consumers should not reference this type; if you have a use case, please file an issue. (#245)
```
liliankasem: This should be formatted:

```
#### Breaking Changes

- `McpInputSchemaJsonUtilities` is now `internal` (was `public`). It was never intended as a public API. Consumers should not reference this type; if you have a use case, please file an issue. (#245)

#### Changes

- Validate required prompt arguments (#244)
- Added output schema support on the tool trigger (#245)
```

src/Microsoft.Azure.Functions.Extensions.Mcp/Trigger/McpToolTriggerAttribute.cs
```diff
@@ -40,6 +40,13 @@ public sealed class McpToolTriggerAttribute(string toolName, string? description
...
+    /// <summary>
+    /// Gets or sets the MCP tool output schema as a JSON string.
+    /// This should only be set when the worker provides an output schema.
+    /// The schema JSON string must be a valid JSON Schema.
+    /// </summary>
+    public string? OutputSchema { get; set; }
+
     /// <summary>
     /// Gets or sets whether the result schema should be provided by the worker instead of being generated by the host extension.
     /// When set to true, the <see cref="ResultSchema"/> property must be provided.
```
liliankasem: Sorry I know this isn't your PR, but there is no `ResultSchema` property. When set to true, you must provide a Mcp_Result type i.e. McpToolResult (or McpPromptResult in future)

The contract UseResultSchema enforces is on the return type, not a separate schema property.

src/Microsoft.Azure.Functions.Extensions.Mcp/DefaultToolRegistry.cs
```diff
@@ -41,6 +41,7 @@ public ValueTask<ListToolsResult> ListToolsAsync(CancellationToken cancellationT
                 Name = tool.Name,
                 Description = tool.Description,
                 InputSchema = tool.GetPropertiesInputSchema(),
+                OutputSchema = tool.OutputSchema,
```
liliankasem, line 44: There's no test for this. A small unit test against DefaultToolRegistry (or the listener's IMcpTool.OutputSchema property) confirming the value reaches the protocol response would close the loop. Right now we only test that we can parse the schema, not that it's actually advertised.

aishwaryabh: added tests for defaultToolRegistry!

kshyju (approved):

liliankasem (approved):

PR #255, opened 2026-04-24
Fix MCP App FileViewSource path resolution on Azure

liliankasem:

MCP App tools configured with `WithView("relative/path.html")` (or `McpViewSource.FromFile`) work locally but fail on Azure Functions (Linux) at invocation time with:

```
System.IO.DirectoryNotFoundException: Could not find a part of the path '/tmp/functions/standby/wwwroot/app/dist/index.html'.
   at System.IO.File.InternalReadAllTextAsync(...)
   at FunctionsMcpAppMiddleware.Invoke(...)
```

Root cause: `FileViewSource` passed the user-provided path straight to `File.ReadAllTextAsync`, which resolves relative paths against `Environment.CurrentDirectory`. On Linux Functions the process working directory is the placeholder/standby location (`/tmp/functions/standby/wwwroot`) and is **not** updated after specialization, so the file is looked up in the wrong place. Locally `cwd == app dir`, so the bug was masked.

**Fix:** resolve relative paths in `FileViewSource` against `AppContext.BaseDirectory` (the deployed app's directory). Absolute paths pass through unchanged. `ToolOptionsValidator` now also fails fast at startup if the resolved file is missing, so users get a clear error instead of a runtime `DirectoryNotFoundException` on first invocation. Doc comments on `WithView(string)` / `McpViewSource.FromFile` now call out the requirement to copy the file to the output directory (e.g., `<Content Include="..." CopyToOutputDirectory="PreserveNewest" />`).

### Pull request checklist

  * [x] Otherwise: Documentation issue linked to PR — XML doc comments updated on the public API surface

### Additional information

Behavior change worth noting: previously a user could rely on cwd-relative paths working locally even when the file wasn't in the build output. After this fix, the file must exist under `AppContext.BaseDirectory` both locally and on Azure (which is the same requirement as deployment). The new validator failure surfaces this clearly at startup. All 500 existing worker tests pass.

(Replaces #254, which had a poorly named branch.)

### DIFF
--- a/src/Microsoft.Azure.Functions.Worker.Extensions.Mcp/Configuration/ToolOptionsValidator.cs
+++ b/src/Microsoft.Azure.Functions.Worker.Extensions.Mcp/Configuration/ToolOptionsValidator.cs
@@ -1,6 +1,7 @@
 // Copyright (c) Microsoft Corporation. All rights reserved.
 // Licensed under the MIT License.
 
+using Microsoft.Azure.Functions.Worker.Extensions.Mcp.McpApp;
 using Microsoft.Extensions.Options;
 
 namespace Microsoft.Azure.Functions.Worker.Extensions.Mcp.Configuration;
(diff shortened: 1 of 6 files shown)


aishwaryabh (approved):
lgtm!

PR #268, opened 2026-06-18
Update Token Utility

satvu:

Add support for CBC token reads on top of GCM as part of the transition for full use of CBC + HMAC.

### Pull request checklist


jviau:
@satvu do you know if this is a behavior breaking change in any way? Can existing customers be broken by this?

satvu:
> @satvu do you know if this is a behavior breaking change in any way? Can existing customers be broken by this?

Per conversation w/ @fabiocav prior to introducing this change, this should not be breaking. The encryption/decryption process here is only on our side (this is basically changing implementation details, on cx end behavior should be the same).

liliankasem:
> > @satvu do you know if this is a behavior breaking change in any way? Can existing customers be broken by this?
> 
> Per conversation w/ @fabiocav prior to introducing this change, this should not be breaking. The encryption/decryption process here is only on our side (this is basically changing implementation details, on cx end behavior should be the same).

Just want to make sure this is captured. There is a "breaking change" here but it is all internal and absorbed. The wire format does change with this PR. This switches the token layout from AES-GCM to AES-CBC + HKDF key derivation, meaning tokens produced by the previous version can't be decrypted by this one. BUT, since consumers must update the NuGet package and redeploy to pick this up, old-format tokens can't reach new-format code outside of the deployment restart window  so there should be no real customer impact.

src/Microsoft.Azure.Functions.Extensions.Mcp/TokenUtility.cs
```diff
@@ -11,87 +11,159 @@ namespace Microsoft.Azure.Functions.Extensions.Mcp;
...
-        using (var hmac = new HMACSHA256(key))
+        catch
         {
-            hmac.TryComputeHash(dataToSign, computedSignature, out _);
+            aes.Dispose();
+            throw;
         }
+    }
+
+    private static byte[] ComputeHmac(byte[] key, ReadOnlySpan<byte> data)
```
liliankasem: nit: ComputeHmac allocates unnecessarily. `.ToArray()` on the ReadOnlySpan<byte> copies to heap. Could use hmac.TryComputeHash() with a span instead.

liliankasem (dismissed):

src/Microsoft.Azure.Functions.Extensions.Mcp/TokenUtility.cs
```diff
@@ -11,87 +11,164 @@ namespace Microsoft.Azure.Functions.Extensions.Mcp;
...
+        var (encKey, macKey) = DeriveKeys(key);
 
-        Span<byte> iv = stackalloc byte[IvSize];
-        RandomNumberGenerator.Fill(iv);
+        byte[] iv = GenerateIv();
+        byte[] ciphertext = Encrypt(Encoding.UTF8.GetBytes(uriState), encKey, iv);
+        byte[] payload = Concat(iv, ciphertext);
+        byte[] signature = ComputeHmac(macKey, payload);
+
+        return WebEncoders.Base64UrlEncode(Concat(payload, signature));
```
fabiocav: The token format changed in a breaking way (IV 12→16, GCM tag removed, new HKDF/HMAC layout) with no version/format byte. Old tokens can't be read by the new code and vice-versa, so any in-flight `state` tokens during a rolling deploy will fail to decrypt. Consider prefixing a 1-byte version so the format can evolve and old tokens can be rejected/migrated explicitly rather than surfacing as a generic crypto failure.

fabiocav: **Backward compatibility — this is the blocking concern.** This PR both (a) stops emitting the old AES‑GCM token format and (b) removes the ability to read it. During a rollout of a serviceable component — especially extension bundles — we'll have a mixed fleet of instances running old and new logic for some period. That breaks in **both** directions:

- An **old** instance receives a **new**-format token → can't decrypt → customer downtime.
- A **new** instance receives an **old**-format token → can't decrypt (old logic was deleted) → failures.

Recommend a **three-phase migration** so no instance ever sees a token it can't handle:

**Phase 1 (this PR should become read-compatible, write-unchanged):**
- Keep emitting the **old** format (existing AES‑GCM encrypt + sign) — do not switch emission yet.
- Add the ability to **read both** formats. Carry a version discriminator so the reader can pick logic: new-format tokens get an explicit version marker; tokens without it are treated as legacy.
- Because legacy tokens have no marker, detect by the new-format marker and, on any new-format verification failure, fall back to legacy parsing (both formats are HMAC/AEAD-authenticated, so a wrong-format guess fails safely rather than silently mis-decrypting). This avoids the ~1/256 ambiguity of a bare leading byte colliding with a random legacy IV byte.

**Phase 2:** Release Phase 1 and wait for full customer adoption, so every instance can *read* the new format.

**Phase 3:** In a subsequent release, flip emission to the **new** format. Safe now, since all instances can already read it.

As written, this PR is Phase 3 without Phases 1–2, which is the unsafe ordering. Suggest reworking it into the Phase 1 shape (dual-read, legacy-write, versioned) and deferring the emission switch to a later release.

fabiocav: **Suggested structure for the phased migration: an `IUriStateProtector` abstraction with one implementation per algorithm.** This keeps both schemes side-by-side, makes read-selection explicit, and turns "remove the old format" into "delete one class."

```csharp
internal interface IUriStateProtector
{
    int MinTokenLength { get; }
    byte[] Protect(string uriState, byte[] key);
    // Returns false (no throw) when the token isn't this format or fails auth,
    // so the caller can try another format.
    bool TryRead(ReadOnlySpan<byte> token, byte[] key, [NotNullWhen(true)] out string? uriState);
}
```

- `AesGcmUriStateProtector` — legacy (v0). Preserves the exact current wire format `[iv12 | ciphertext | tag16 | signature32]` so existing tokens still decode. Unmarked ⇒ it's the catch-all reader, tried last.
- `AesCbcHmacUriStateProtector` — new (v1). This PR's scheme plus a leading version byte: `[0x01 | iv16 | ciphertext | signature32]`, HKDF-derived enc/mac keys, HMAC over `version|iv|ciphertext`, verified before decrypt.

`TokenUtility` becomes a thin orchestrator:

```csharp
private static readonly IUriStateProtector Writer = LegacyProtector; // Phase 1: still emit legacy
private static readonly IUriStateProtector[] Readers = { NewProtector, LegacyProtector };
private static readonly int MinTokenLength = Readers.Min(r => r.MinTokenLength);

public static string ProtectUriState(string s, byte[]? key = null)
    => WebEncoders.Base64UrlEncode(Writer.Protect(s, key ??= GetKey()));

public static string ReadUriState(string token, byte[]? key = null)
{
    key ??= GetKey();
    var bytes = WebEncoders.Base64UrlDecode(token);
    if (bytes.Length < MinTokenLength)
        throw new InvalidOperationException("Token is too short or malformed.");
    foreach (var r in Readers)
        if (r.TryRead(bytes, key, out var state)) return state;
    throw new CryptographicException("Invalid token signature.");
}
```

Why `TryRead`-with-fallback rather than selecting purely on the version byte: legacy tokens have no marker and their first byte is a random IV byte, so a bare "first byte == 0x01 ⇒ new format" check would misroute ~1/256 legacy tokens and reject them. Because every format is authenticated, trying the version-marked reader first and falling back on a verification miss is safe and unambiguous.

Migration then reduces to:
- **Phase 1 (this PR):** `Writer = Legacy`, `Readers = { New, Legacy }` — emit old, read both.
- **Phase 2:** ship and wait for full adoption.
- **Phase 3:** `Writer = New`. Later, delete `AesGcmUriStateProtector` and drop it from `Readers`.

I validated this shape locally (existing tests plus new-format round-trip, cross-format reads, tamper rejection, and a guard asserting the writer still emits legacy) — all green, then reverted so the PR stays as-is per the plan to leave this as guidance.

jviau: @fabiocav, while it would be great to do a phased rollout of this, I don't know if we have the ability to do that reliably. Dotnet customers are in full control of their extensions, so it would be completely on them to know to update their app in a phased fashion.

Additionally, bundles customers can be specifying bundles and jump past the initial phases of this rollout. Would this also be affected by any abandoned or reverted bundle releases?

So while it is the right thing to avoid any downtime, given the lack of guarantees around it, can we even execute the phased plan? Is it still worth the effort?

src/Microsoft.Azure.Functions.Extensions.Mcp/Security/AesCbcHmacUriStateProtector.cs
```diff
@@ -0,0 +1,136 @@
...
+    {
+        uriState = null;
+
+        if (token.Length < MinTokenLength)
+        {
+            return false;
+        }
+
+        // Check version marker
+        if (token[0] != Version)
```
jviau, line 62: Does the legacy token format have any chance of the first byte coincidentally matching `Version` here?

Looking at it, yes there is as the first byte is completely random on the legacy format. The next question is if this would successfully decrypt a legacy format (into junk data), or will it fail and let this pass on to the legacy implementation?

src/Microsoft.Azure.Functions.Extensions.Mcp/Security/AesCbcHmacUriStateProtector.cs
```diff
@@ -0,0 +1,136 @@
...
+        {
+            aes.Key = key;
+            aes.IV = iv;
+            aes.Mode = CipherMode.CBC;
+            aes.Padding = PaddingMode.PKCS7;
+            return aes;
+        }
+        catch
+        {
+            aes.Dispose();
```
jviau, line 133: Add a comment for clarity "Dispose only on failure."

jviau (commented):
2 primary concerns:

1. Is the phased approach even worth the effort given how we can't accurately control a phased rollout here, ensuring customers only move one phase at a time.
2. What happens when the legacy format happens to match the new formats `Version` byte marker?

src/Microsoft.Azure.Functions.Extensions.Mcp/Security/AesCbcHmacUriStateProtector.cs
satvu: There is a chance, but there's a test for it: `ReadUriState_LegacyTokenStartingWith0x01_FallsBackToGcmReader`. The flow is: 

When a legacy token's first byte happens to be 0x01:

- TokenUtility.ReadUriState iterates readers: [AesCbcHmacUriStateProtector, AesGcmUriStateProtector]
- CBC reader tries first → sees token[0] == 0x01 (version match!) → proceeds to HKDF-derive enc/mac subkeys → computes HMAC over the payload using those derived keys → HMAC mismatch (legacy tokens were signed with the raw key, not HKDF-derived keys) → returns false
- GCM reader tries next → verifies HMAC with raw key → succeeds → decrypts via AES-GCM → returns the plaintext ✅

satvu:
> 1. Is the phased approach even worth the effort given how we can't accurately control a phased rollout here, ensuring customers only move one phase at a time.

The phased rollout would at least minimize failures by giving instances time to pick up new bundles/app versions with the ext update on restart. We can't ensure all customers will have picked this up before the next phase(s), but I think _enough_ will to avoid a major incident.

jviau (dismissed):

jviau (approved):

satvu:
\azp run

